PR #33, opened 2017-04-02
Generator: haskell: return malloc data instead of local static

Dettorer:
This allows to call cxx2lang_array_ptr and cxx2lang multiple times
without overwriting a previous array.

This also makes these functions not a unique symbol anymore and allows
haskell's interface.cc to be compiled using the SYSV ABI. Unique symbols
are a GNU extension, using them forces us to compile using the GNU/Linux
ABI while we compile everything else with SYSV ABI.


seirl:
<3

PR #37, opened 2018-05-07
Add Rust support

MaloJaffre:
This reuses part of the C generator, by using a C++ <=> C <=> Rust FFI.
The linking is done with whole-archive to properly merge C/C++ and Rust static libraries.

`testall.sh` succeeds on my machine, and I've manually verified the correctness of the generated `prologin2017` environment.
Related to prologin/sadm#102.

### DIFF
--- a/tools/generator/files/rules.mk
+++ b/tools/generator/files/rules.mk
@@ -30,6 +30,7 @@ ifndef NOCOLORS
   quiet_cmd_ocamlo    = [1;33mcaml   $< -> $@[0m
   quiet_cmd_hsc2hs    = [1;32mhsc2hs $< -> $@[0m
   quiet_cmd_ghc       = [1;33mghc    $^ -> $@[0m
+  quiet_cmd_rustc     = [1;33mrustc  $^ -> $@[0m
   quiet_cmd_ld_shared = [1;36mlib    $@[0m
   quiet_cmd_clean     = [1;35mclean[0m
   quiet_cmd_distclean = [1;35mdistclean[0m
@@ -43,6 +44,7 @@ else
   quiet_cmd_ocamlo    = OCAML   $@
   quiet_cmd_hsc2hs    = HSC2HS  $@
   quiet_cmd_ghc       = GHC     $@
+  quiet_cmd_rustc     = RUSTC   $@
   quiet_cmd_ld_shared = LINK    $@
   quiet_cmd_clean     = CLEAN   objects
   quiet_cmd_distclean = CLEAN   targets
@@ -66,6 +68,7 @@ JAVAC  = javac
 OCAMLC = $(CROSS)ocamlc
 GHC    = $(CROSS)ghc
 HSC2HS = $(CROSS)hsc2hs
+RUSTC  = $(CROSS)rustc
 LD     = $(CROSS)ld
 
 OCAML_LIBS     = -L`ocamlc -where` -Wl,-R`ocamlc -where` -lcamlrun_shared -lcurses -lm
@@ -153,6 +156,7 @@ define build_lib
   $(1).so: override _CPPFLAGS = $$($(1)-cppflags)
   $(1).so: override _CFLAGS = $$($(1)-cflags)
   $(1).so: override _CXXFLAGS = $$($(1)-cxxflags)
+  $(1).so: override _RUSTCFLAGS = $$($(1)-rustcflags)
   ifdef $(1)-deps
     $(1).so: $$($(1)-deps)
   endif
@@ -162,21 +166,24 @@ $(1).so: $$(_obj)
 endef
 
 
-c_flags    = $(_CFLAGS) -MD -MP -MF $(@D)/.$(@F).d
-cxx_flags  = $(_CXXFLAGS) -std=c++14 -MD -MP -MF $(@D)/.$(@F).d
-cpp_flags  = $(_CPPFLAGS)
-cmd_cc     = $(CC) $(c_flags) $(cpp_flags) -fPIC -c $< -o $@
-cmd_cxx    = $(CXX) $(cxx_flags) $(cpp_flags) -fPIC -c $< -o $@
-cmd_java   = $(JAVAC) $(java_flags) $<
-cmd_ocaml  = $(OCAMLC) $(_CAMLFLAGS) -c $< -o $@
-cmd_ocamlo = $(OCAMLC) -output-obj $(_CAMLFLAGS) $(filter %.cmo,$^) -o $@
-cmd_hsc2hs = $(HSC2HS) $< -o $@
-cmd_ghc    = $(GHC) -pgml $(CXX) -O9 -dynamic --make -shared -fPIC \
-             -L`$(GHC) --print-libdir` -lHSrts-ghc`$(GHC) --numeric-version` \
-             $(filter %.hs %.o %.a,$^) -o $@
+c_flags     = $(_CFLAGS) -MD -MP -MF $(@D)/.$(@F).d
+cxx_flags   = $(_CXXFLAGS) -std=c++14 -MD -MP -MF $(@D)/.$(@F).d
+cpp_flags   = $(_CPPFLAGS)
+rustc_flags = $(_RUSTCFLAGS) -C relocation-model=pic -C panic=abort \
+              --crate-type=staticlib --emit link,dep-info=$(@D)/.$(@F).d
+cmd_cc      = $(CC) $(c_flags) $(cpp_flags) -fPIC -c $< -o $@
+cmd_cxx     = $(CXX) $(cxx_flags) $(cpp_flags) -fPIC -c $< -o $@
+cmd_java    = $(JAVAC) $(java_flags) $<
+cmd_ocaml   = $(OCAMLC) $(_CAMLFLAGS) -c $< -o $@
+cmd_ocamlo  = $(OCAMLC) -output-obj $(_CAMLFLAGS) $(filter %.cmo,$^) -o $@
+cmd_hsc2hs  = $(HSC2HS) $< -o $@
+cmd_ghc     = $(GHC) -pgml $(CXX) -O9 -dynamic --make -shared -fPIC \
+              -L`$(GHC) --print-libdir` -lHSrts-ghc`$(GHC) --numeric-version` \
+              $(filter %.hs %.o %.a,$^) -o $@
+cmd_rustc   = $(RUSTC) $(rustc_flags) $< -o $@
 
 ld_flags      = $(_LDFLAGS)
-cmd_ld_shared = $(CXX) $(filter %.o %.a,$^) $(ld_flags) -shared -o $@ $(_LDLIBS)
+cmd_ld_shared = $(CXX) $(ld_flags) $(filter %.o %.a,$^) $(_LDLIBS) -shared -o $@
 
 cmd_clean     = $(RM) $(_cleanfiles)
 cmd_distclean = $(RM) $(_dcleanfiles)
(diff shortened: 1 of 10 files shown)


zopieux:
Woaw, that's a nice contribution, thanks! We'll look into it as soon as possible.

MaloJaffre:
I've now updated this PR with a new generic struct `Array<T>` and a method `free` to handle arrays more ergonomically than with the previous arrays `type_array` of the C API.

tools/generator/files/rules.mk
```diff
@@ -162,21 +166,30 @@ $(1).so: $$(_obj)
...
+cmd_hsc2hs  = $(HSC2HS) $< -o $@
+cmd_ghc     = $(GHC) -pgml $(CXX) -O9 -dynamic --make -shared -fPIC \
+              -L`$(GHC) --print-libdir` -lHSrts-ghc`$(GHC) --numeric-version` \
+              $(filter %.hs %.o %.a,$^) -o $@
+cmd_rustc   = $(RUSTC) $(rustc_flags) $< -o $@
 
 ld_flags      = $(_LDFLAGS)
-cmd_ld_shared = $(CXX) $(filter %.o %.a,$^) $(ld_flags) -shared -o $@ $(_LDLIBS)
+ifeq ($(STECHEC_LANG),rust)
+  cmd_ld_shared = $(CXX) -Wl,--whole-archive $(filter %.o %.a,$^) \
```
seirl: Can't all of those go directly into $ldflags?

tools/generator/gen/generator_c.rb
```diff
@@ -232,8 +232,8 @@ def generate_header
     @f.close
   end
 
-  def build
-    @path = Pathname.new($install_path) + "c"
+  def build(lang = "c")
+    @path = Pathname.new($install_path) + lang
```
seirl: I agree that the design here isn't ideal, but I think it would still be a bit more elegant to subclass CCxxFileGenerator and override build() there.

tools/generator/test/champions/prologin.rs
```diff
@@ -0,0 +1,132 @@
+#![allow(bad_style, unused)]
+mod api;
+use api::*;
+use std::os::raw::{c_double, c_int, c_void};
+use std::slice::from_raw_parts;
+
+/// Called 10K times to test if things work well.
+#[no_mangle]
+pub unsafe extern "C" fn test()
```
seirl: *Ideally* I'd think it would be great to have a wrapper that does the extern "C" and the demangling, and those wrappers would call the user functions, which I think would be a bit prettier (and less boilerplate for the contestant). If you don't have time to do it, that's fine, but that would be a good improvement.

seirl:
I added a few comments of things that could be better. Just so you know, we're definitely planning on testing that for next week's finals.

tools/generator/files/rules.mk
MaloJaffre: The problem is the `-Wl,--whole-archive` argument which is required to be before the libraries linked.
I could add a `pre-ld-flags` variable to generalize this, but I'm not very proficient in Makefiles writing, and this is a very special case of `cmd_ld_shared`.

seirl: I think LDFLAGS is supposed to be before the objects, and what goes after is in LDLIBS. Can you still compile the other champions if you just put $ldflags before $(filter...)?

MaloJaffre:
@seirl I think I've now addressed all the comments made so far.

Diff with previous version:  
```diff
diff --git a/tools/generator/files/rules.mk b/tools/generator/files/rules.mk
index 4e77b20..914ce32 100644
--- a/tools/generator/files/rules.mk
+++ b/tools/generator/files/rules.mk
@@ -183,13 +183,7 @@ cmd_ghc     = $(GHC) -pgml $(CXX) -O9 -dynamic --make -shared -fPIC \
 cmd_rustc   = $(RUSTC) $(rustc_flags) $< -o $@
 
 ld_flags      = $(_LDFLAGS)
-ifeq ($(STECHEC_LANG),rust)
-  cmd_ld_shared = $(CXX) -Wl,--whole-archive $(filter %.o %.a,$^) \
-                  -Wl,--no-whole-archive -lm -lrt -ldl -lpthread -lstdc++ \
-                  -shared -fPIC -o $@ $(_LDLIBS)
-else
-  cmd_ld_shared = $(CXX) $(filter %.o %.a,$^) $(ld_flags) -shared -o $@ $(_LDLIBS)
-endif
+cmd_ld_shared = $(CXX) $(ld_flags) $(filter %.o %.a,$^) $(_LDLIBS) -shared -o $@
 
 cmd_clean     = $(RM) $(_cleanfiles)
 cmd_distclean = $(RM) $(_dcleanfiles)
diff --git a/tools/generator/gen/generator_c.rb b/tools/generator/gen/generator_c.rb
index 0a397ce..b7cc5a1 100644
--- a/tools/generator/gen/generator_c.rb
+++ b/tools/generator/gen/generator_c.rb
@@ -232,8 +232,8 @@ EOF
     @f.close
   end
 
-  def build(lang = "c")
-    @path = Pathname.new($install_path) + lang
+  def build
+    @path = Pathname.new($install_path) + "c"
     @source_file = 'interface.cc'
     @header_file = 'interface.hh'
 
diff --git a/tools/generator/gen/generator_rust.rb b/tools/generator/gen/generator_rust.rb
index 72c9e16..863cedf 100644
--- a/tools/generator/gen/generator_rust.rb
+++ b/tools/generator/gen/generator_rust.rb
@@ -35,6 +35,17 @@ def rust_proto(fn)
   "fn #{fn.name}(#{args})" + ret
 end
 
+class CCxxFileGeneratorForRust < CCxxFileGenerator
+  def build
+    @path = Pathname.new($install_path) + "rust"
+    @source_file = 'interface.cc'
+    @header_file = 'interface.hh'
+
+    generate_source
+    generate_header
+  end
+end
+
 class RustFileGenerator < FileGenerator
   def print_multiline_comment(str, prestr = '')
     str.each_line {|s| @f.print prestr, "// ", s }
@@ -116,18 +127,23 @@ EOF
       @f.print "pub ", rust_proto(f), ";\n"
     end
     @f.puts "}", ""
+    for_each_user_fun do |f|
+      @f.puts"#[no_mangle]"
+      @f.print "pub unsafe extern \"C\" ", rust_proto(f)
+      args = f.args.map { |arg| arg.name }.join(", ")
+      print_body "  super::#{f.name}(#{args})"
+    end
     @f.close
   end
 
   def generate_user
     @f = File.open(@path + @user_file, 'w')
     @f.puts "#![allow(unused)]"
-    @f.puts "mod api;"
+    @f.puts "pub mod api;"
     @f.puts "use api::*;"
     @f.puts "use std::os::raw::{c_double, c_int, c_void};", ""
     for_each_user_fun do |f|
-      @f.puts"#[no_mangle]"
-      @f.print "pub extern \"C\" ", rust_proto(f)
+      @f.print rust_proto(f)
       print_body "  // fonction a completer"
     end
     @f.close
@@ -148,6 +164,9 @@ lib_TARGETS = #{target}
 # Evite de toucher a ce qui suit
 #{target}-dists += #{$conf['conf']['player_filename']}.h interface.hh #{@api_file}
 #{target}-srcs += interface.cc #{@user_file}
+#{target}-ldflags = -fPIC -Wl,--whole-archive
+#{target}-ldlibs  = -Wl,--no-whole-archive -lm -lrt -ldl -lpthread -lstdc++
+
 STECHEC_LANG=rust
 include ../includes/rules.mk
     EOF
@@ -156,7 +175,7 @@ include ../includes/rules.mk
 
 
   def build()
-    CCxxFileGenerator.new.build("rust")
+    CCxxFileGeneratorForRust.new.build
 
     @path = Pathname.new($install_path) + "rust"
     @user_file = $conf['conf']['player_filename'] + '.rs'
diff --git a/tools/generator/test/champions/prologin.rs b/tools/generator/test/champions/prologin.rs
index 2dee6b5..4440852 100644
--- a/tools/generator/test/champions/prologin.rs
+++ b/tools/generator/test/champions/prologin.rs
@@ -1,12 +1,11 @@
 #![allow(bad_style, unused)]
-mod api;
+pub mod api;
 use api::*;
 use std::os::raw::{c_double, c_int, c_void};
 use std::slice::from_raw_parts;
 
 /// Called 10K times to test if things work well.
-#[no_mangle]
-pub unsafe extern "C" fn test()
+unsafe fn test()
 {
     assert!(CONST_VAL/4 == 10);
     assert!(CONST_DOUBLE_2/2. == 668.5);

tools/generator/files/rules.mk
MaloJaffre: I've tested and it works for every language, except php and haskell, because I've not managed to build them.

MaloJaffre: I've tested and it works for every language, except php and haskell, because I've not managed to build them.

seirl:
Thanks, looking into that. Btw, I'd much rather have multiple commits that we squash when merging the PR than diffs in the comments ;-)

seirl:
I think you're missing the makefile_rust.rb?

MaloJaffre:
@seirl Thank you very much for your review, and I'll try not to squash commits in the future.

It seems that `makefile_c.rb` is not used for example for generating a client in C, because the code is directly in `generator_c.rb`, but I maybe miss something?

seirl:
Yeah, there are two different makefiles that we need to have. The one you're refering to is used for local matches, but the server needs `makefile_XXX` to generate its own makefile: https://github.com/prologin/stechec2/blob/e1c6e613fde89e4f97eb5194ab16262640e83b95/tools/generator/generator.rb#L119-L145

MaloJaffre:
@seirl Thanks for this hint, I've now pushed a new commit adding that file, but I was unable to find a way to test it.

seirl:
This looks good, I'll try to test that for you from my local setup, don't worry.

seirl:
Merging to ease internal testing, we'll iterate from this if we find any issues.

MaloJaffre:
@seirl Thanks again for your review!

Namarand:
@MaloJaffre All API functions are currently unsafe, making them quite painful to use. One solution is to wrap all function calls in an unsafe block, allowing the user to use directly the functions. Does this look good to you? If yes, I'll do it asap.

MaloJaffre:
@Namarand Good idea!
Everything in the API is actually safe, excepting `free` and all function that takes an `Array<T>` (`ptr` must be pointing to a valid allocation of `len` eleements).
Apart from this detail, this wrapping looks good to me.
You could rename the `api` module to `ffi` and put the nice wrapping in `api` to cleanly separate the exposed API from the internal implementation.

seirl:
@malojaffre I really don't like the Array implementation. I think the wrapper function should copy the array in a native rust array, we really don't want a free().

MaloJaffre:
@seirl @Namarand I don't like the rawness of the exposed API neither, so I will send a PR to add a nice wrapping (with `ffi` and `api` modules) asap.

seirl:
@MaloJaffre I think the main problem is that you used the C api, so instead of doing a nice conversion C++ -> Rust it does C++ -> C -> Rust, so you have to go through all the hacks that are made just to have usable arrays in C. I would like it a lot more if you could just populate your Rust arrays from the C++ vectors. Do you think that's doable?

MaloJaffre:
@seirl I will try to do it by directly allocating a Rust `Vec<T>` in C++, but the interface will still pass trough `extern "C"` functions because C++ mangling is not implemented in `rustc`

seirl:
It's not a problem to pass through extern "C" functions, but I just don't think you should use the C generator. The C++ generator already contains those unmangled function exports.

Namarand:
Actually, https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html does quite a good job for converting C array to slice. Maybe we can use that ? We can also convert them to Vec<> thanx to the to_vec() method.

MaloJaffre:
@seirl  I will rework the API to avoid the use of the C generator asap.
@Namarand Good idea, I will definitely use this function in the rewrite.

PR #39, opened 2018-05-15
Wrap the Rust FFI in an idiomatic API

MaloJaffre:
As discussed in #37.

This doesn't yet eliminate the C layer, because it seems very difficult 
to allocate and construct in C++ a `Vec<T>` (its layout is unspecified so
that its fields can be reordered by rustc, and it is incompatible with malloced
buffers), but all these horrible `Array<T>` are now hidden in a new `ffi`
module.

Nevertheless, this greatly simplifies the API usage, as seen in the
`prologin.rs` test, even if it is not the cleanest solution with regards
to the internal implementation of the FFI.

cc @seirl @Namarand.

### DIFF
--- a/tools/generator/files/useapi.tex
+++ b/tools/generator/files/useapi.tex
@@ -118,12 +118,5 @@ fonctionACompleter = do
 \item{Les noms des structures et énumérations sont en \texttt{CamelCase}.
       Ainsi, une structure nommée \texttt{foo\_bar} dans le sujet
       s'appellera \texttt{FooBar} en Rust.}
-\item{Les fonctions prenant des tableaux en paramètres et retournant des
-      tableaux utilisent à la place de ces tableaux une structure
-      \texttt{Array<T>}, où \texttt{T} est le type des données dans le
-      tableau. Ces structures contiennent deux éléments : les données,
-      \texttt{ptr: *const T}, et la taille, \texttt{len: usize}.
-      Dans tous les cas, la libération des données est laissée au soin du
-      candidat, en utilisant la méthode \texttt{unsafe fn free(self: Array<T>);}
-      fournie. ;}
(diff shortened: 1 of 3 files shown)


seirl:
Much better!

> This doesn't yet eliminate the C layer, because it seems very difficult
> to allocate and construct in C++ a Vec<T> (its layout is unspecified so
> that its fields can be reordered by rustc, and it is incompatible with malloced
> buffers), but all these horrible Array<T> are now hidden in a new ffi
> module.

Suggestion to fix that: export in the rust interface a non-mangled function that creates a Vec from a C array and a length, and one that fills a C array and a length pointer from a Vec, then just call those demangled functions from lang2cxx_array and cxx2lang_array? I think that would allow us to ditch completely that C interface layer.

seirl:
By the way, it's interesting to keep those iterations in the git history, so feel free to just add more commits to this branch if you fix this.

seirl:
@MaloJaffre do you think you have time to work on the change we talked about earlier? Ideally we'd like to print the paper subjects as soon as possible for this final, so it would be cool to have a final version of the api interface in the .tex.

MaloJaffre:
@seirl I think the changes remaining will only affect the internal implementation (the FFI between C++ and Rust), and not the exposed interface in `api.rs`, so this PR is reviewable now, because I am not planning further improvements to the exposed API.

As for eliminating this C layer, it turns out to be really hard, because structs containing Vecs cannot be `#[repr(c)]` and so, if we eliminate the C intermediary structs, C++ need to pass structs fields to the Rust side which returns the constructed struct, and I feel this is as complicated as the current situation.

seirl:
Ok, let's merge this for now and we can keep discussing the data interchange format after in a separate issue.

PR #43, opened 2019-05-25
Upgrade to Rust edition 2018

MaloJaffre:
This brings various improvements to the language, like C++17.

The small changes are necessary to pass the tests.

### DIFF
--- a/tools/generator/files/rules.mk
+++ b/tools/generator/files/rules.mk
@@ -169,7 +169,7 @@ endef
 c_flags     = $(_CFLAGS) -MD -MP -MF $(@D)/.$(@F).d
 cxx_flags   = $(_CXXFLAGS) -std=c++14 -MD -MP -MF $(@D)/.$(@F).d
 cpp_flags   = $(_CPPFLAGS)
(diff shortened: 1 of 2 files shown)


seirl:
We strive to be compatible with debian stable. Support for C++17 is a bit primitive in GCC6. Is this flag supported in the latest debian stable version of rustc?

MaloJaffre:
This flag isn't supported in debian stable (rustc 1.24.1), but Rust is evolving rapidly and there is a lag of 11 versions compared to archlinux (rustc 1.35.0), with significant improvements for the language, especially with this flag enabled.

seirl:
Hmm, right... I'm willing to make an exception because we're so close to the Debian Buster release, and this is isolated in a generator, but it's really annoying for development to not be able to use Stechec2 when you're not on Archlinux.

One other possibility in the future could be to override the RUSTC options in the contestant Makefile?

alarsyo:
According to https://packages.debian.org/buster/rustc this is no longer an issue (Support for the 2018 edition was [stabilized in Rust 1.31](https://blog.rust-lang.org/2018/12/06/Rust-1.31-and-rust-2018.html), and Debian stable now ships `rustc` v1.34)

PR #51, opened 2019-06-10
Apply clang-format

audebert:
The clang-format configuration was tweaked to match the current style.

The biggest source of change is that clang-format does not support the preprocessor indent that was used.

Next step is to enforce style in PR hook.


seirl:
> The biggest source of change is that clang-format does not support the preprocessor indent that was used.

Can't we just use `#pragma once` everywhere anyway?

seirl (approved):

audebert:
Let's use `#pragma once` in a second PR.

PR #52, opened 2019-06-10
Add copyright notice

audebert:


Shaac (approved):
I’m not sure we have the right to copyright some old code with the year 2019.
The code was licensed several years ago already, so maybe we have to use a year like 2012-2019.

audebert:
> I’m not sure we have the right to copyright some old code with the year 2019.
> The code was licensed several years ago already, so maybe we have to use a year like 2012-2019.

Good point, I've set the copyright to the year the file was added, which reflects when we first can first claim copyright on it.

PR #53, opened 2019-06-10
Refactor GameState using new GameStateHistory

audebert:
Changes:

- No more naked pointers to GameState, only `std::unique_ptr`s. This simplifies
  the code and make mistakes harder to make.
- rules::Action do not handle GameState anymore, they only have to know how to
  mutate the GameState they are given.
- Api is now responsible to call `game_state_.save()` before applying each
  action. This makes a copy of the current GameState and store it to be used if
  cancel() is called.
- Rules do not have GameState anymore, because they cannot use cancel().
- GameState object are held by a utils::VersionedPtr container, which knows how
  to save and restore GameState instances.

Performance: overall equivalent (same amount of GameState copies), better when
the caller knows that they do not have to support cancel() (no copies are made
if they don't call `game_state.save()`)

Also:

- Replace typedef with using


tools/generator/files/rules.cc
```diff
@@ -1,9 +1,13 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 #include "rules.hh"
+
+#include <memory>
+#include <utility>
+
 #include "actions.hh"
 
 Rules::Rules(const rules::Options opt)
-    :!!rules_cc_inheritance!! sandbox_(opt.time)
```
seirl, line 7: Is that intended? I imagine this would break the template replacement.

seirl (changes requested):

tools/generator/files/rules.cc
audebert: This was not intentional. Thanks.

audebert:
I've been thinking of making a specialized implementation of utils::VersionedPtr named GameStateHistory that allow re-introducing the GameState::copy() method and allow hiding the copy constructor. I'll update the PR with that change.

seirl (approved):

Shaac (approved):

audebert:
Now updating the games. Once this is done, will update everything at once.

PR #54, opened 2019-06-11
Api and GameState interface refactor

audebert:
This PR builds on top of the changes in #53 and is intended to reduce the amount of boilerplate code in the games.

It moves the code of the `Api` class that was generated to a new `rules::Api` class that has the same functionality. This provides a `Api::cancel()` method to all game Apis, but games do not have to export it.

The next refactor in line is to probably to check whether we need really need an `Api` class and if `Rules` should not own the GameState.

### DIFF
--- a/games/tictactoe/src/action_play.hh
+++ b/games/tictactoe/src/action_play.hh
@@ -16,7 +16,9 @@ public:
     ActionPlay() {} // for register_action()
 
     int check(const GameState& st) const override;
-    void apply(GameState& st) const override;
+    void apply_on(GameState* st) const override;
+    // Unhide base class apply()
+    using rules::Action<GameState>::apply;
 
     void handle_buffer(utils::Buffer& buf) override
     {
--- a/src/lib/rules/action.hh
+++ b/src/lib/rules/action.hh
@@ -44,15 +44,18 @@ public:
 template <typename TState>
 class Action : public IAction
 {
+    static_assert(std::is_base_of<rules::GameState, TState>::value,
+                  "TState not derived from rules::GameState");
+
 public:
     virtual int check(const TState& st) const = 0;
     int check(const GameState& st) const override
     {
-        return check(dynamic_cast<const TState&>(st));
+        return check(static_cast<const TState&>(st));
     }
 
     virtual void apply_on(TState* st) const = 0;
-    void apply(GameState* st) const { apply_on(dynamic_cast<TState*>(st)); }
+    void apply(GameState* st) const { apply_on(static_cast<TState*>(st)); }
     void apply(std::unique_ptr<TState>& st) const { apply_on(st.get()); }
 };
 
--- a/src/lib/rules/action.hh
+++ b/src/lib/rules/action.hh
@@ -8,8 +8,6 @@
 
 #include <utils/buffer.hh>
 
-#include "game-state.hh"
-
 namespace rules {
 
 class GameState;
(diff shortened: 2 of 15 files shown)


src/lib/rules/action.hh
```diff
@@ -24,7 +24,7 @@ public:
     virtual int check(const GameState& st) const = 0;
 
     // Apply action to game state.
-    virtual void apply(GameState& st) const = 0;
+    virtual void apply(GameState* st) const = 0;
```
Shaac, line 27: Why changing again? You changed the pointer into a ref in the last PR. And now it is back to pointer.

Personally I strongly prefer references because, pointers may imply ownership, and memory management. But some people have other reasons and prefer pointers.

I’m just puzzled to see it change again. Plus, it is not consistent with the `check` anymore (one uses a ref, the other a pointer).

games/tictactoe/src/action_play.hh
```diff
@@ -16,7 +16,9 @@ public:
     ActionPlay() {} // for register_action()
 
     int check(const GameState& st) const override;
-    void apply(GameState& st) const override;
+    void apply(GameState* st) const override;
+    // Unhide base class apply()
+    using rules::Action<GameState>::apply;
```
Shaac, line 21: I’m not sure what you want to unhide here. The `apply` method is overiden. There is no other `apply` method. Unless I missed something.

Shaac: Ah OK I’m seen the comment on the other file. But this is a bit strange. Hopefully this is some generated code. But can't we use different names?

Shaac: Could be something like : the overridden method in protected. It is called like `apply_`. And the public `apply` method can have tons of overload if you think its usefull (shared_ptr, unique_ptr, ref, ...)

src/lib/rules/action.hh
audebert: I reverted to using a pointer when I started refactoring code that uses this interface.

The rationale is that we should generally not have non-const reference arguments. This is confusing for the user of the interface, especially when you use check() and apply() close together, for example:

```
GameState st{};
ActionFoo action{};
action.check(st); // (1)
action.apply(st); // (2)
```

(1) All good, value-semantic for st, therefore st is a read-only argument, the reader can rightfully expect check() to take a const&

(2) Now in this line st is used with the same value syntax as (1), but the semantic is different: st in apply() is an output argument therefore it must have a reference semantic. The reader does not see the difference and does expect st to be modified. This is bad.

Now with a pointer the ambiguity is resolved by syntax:

```
GameState st{};
ActionFoo action{};
action.check(st); // value-semantic
action.apply(&st); // reference-semantic
```

As for ownership of the pointer in the method impl., we should use std::unique_ptr as the cannonical pointer ownership (and mem management, but that's the job of unique_ptr) marker. Raw pointers should only be used to convey the reference semantic.

games/tictactoe/src/action_play.hh
audebert: Actually, "unhide" is not the right word.

The method is apparently not looked for by the compiler unless we explicitely surface it in the descendant class, and it has nothing to do with the overloads. This is similar to the behavior described in this article: http://www.cs.technion.ac.il/users/yechiel/c++-faq/nondependent-name-lookup-members.html

I'll reword the comments to make them clearer.

I'll check if those files can be generated.

audebert: Did a quick test: the overload is indeed hiding the template-dependent base class method.

The solution you suggested (split overload set and implem method) is actually what is currently done, with the `apply` (overload set) and `apply_on` (implem) methods. This solution is definitvely simpler than the `using` trick. I'll revert this change.

audebert: Updated PR, keep the apply_on() method for impl. and apply() for overloads.

PR #61, opened 2019-07-03
Add plusminus, a synchronous, map based game

audebert:
Used for testing.


audebert:
Any objection to merging this?

PR #62, opened 2019-07-03
Add stechec2-replay

audebert:
From the proposed stechec2-replay documentation:

---

Debugging or testing the rules require starting a server and multiple clients.
This process is simplified by the ``stechec2-run`` utility, but it still
require you to have the champions compiled locally. This setup is not
straightforward and may not even reproduce a specific bug if the champions
generate random actions.

To create a standalone copy of a match, you can create a match replay file with
the ``--replay`` command line argument of ``stechec2-server``:

    stechec2-server --replay match.replay ...

The replay file can be read by the ``stechec2-replay`` utility, which only
requires the rules shared library to be present to replay a match:

    stechec2-replay --replay match.replay --rules rules.so

You can use this feature to replay a match that triggers a specific bug, or use
the replay file a unit test.

---

Next steps:

* Add support for fetching replays in concours/
* Add a text-based action representation for text-based replays
* Add test infrastructure to use replays as test cases.

### DIFF
--- a/doc/rules.rst
+++ b/doc/rules.rst
@@ -726,7 +726,7 @@ requires the rules shared library to be present to replay a match::
 The replay file contains the map, if the game uses one, the list of actions
 sent by the clients (champions and spectators) and the final score.  At the end
 of the replayed match, ``stechec2-replay`` checks if the obtained score matches
-the score save in the reply file.
+the score saved in the reply file.
 
 You can use this tool to replay a match that triggers a specific bug, or use
-the replay file a unit test.
+the replay file as a unit test.
--- a/doc/rules.rst
+++ b/doc/rules.rst
@@ -707,9 +707,9 @@ And that's it!
 Debugging
 =========
 
-Debugging or testing the rules require starting a server and multiple clients.
+Debugging or testing the rules requires starting a server and multiple clients.
 This process is simplified by the ``stechec2-run`` utility, but it still
-require you to have the champions compiled locally. This setup is not
+requires you to have the champions compiled locally. This setup is not
 straightforward and may not even reproduce a specific bug if the champions
 generate random actions.
 
@@ -723,5 +723,10 @@ requires the rules shared library to be present to replay a match::
 
    stechec2-replay --replay match.replay --rules rules.so
 
-You can use this feature to replay a match that triggers a specific bug, or use
+The replay file contains the map, if the game uses one, the list of actions
+sent by the clients (champions and spectators) and the final score.  At the end
+of the replayed match, ``stechec2-replay`` checks if the obtained score matches
+the score save in the reply file.
+
+You can use this tool to replay a match that triggers a specific bug, or use
 the replay file a unit test.
--- a/doc/rules.rst
+++ b/doc/rules.rst
@@ -703,3 +703,25 @@ functions but for round specific needs, such as ``start_of_round``,
 ``end_of_round``, etc.
 
 And that's it!
+
+Debugging
+=========
+
+Debugging or testing the rules require starting a server and multiple clients.
+This process is simplified by the ``stechec2-run`` utility, but it still
+require you to have the champions compiled locally. This setup is not
+straightforward and may not even reproduce a specific bug if the champions
+generate random actions.
+
+To create a standalone copy of a match, you can create a match replay file with
+the ``--replay`` command line argument of ``stechec2-server``::
+
+   stechec2-server --replay match.replay ...
+
+The replay file can be read by the ``stechec2-replay`` utility, which only
+requires the rules shared library to be present to replay a match::
+
+   stechec2-replay --replay match.replay --rules rules.so
+
+You can use this feature to replay a match that triggers a specific bug, or use
+the replay file a unit test.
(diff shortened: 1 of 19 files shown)


doc/rules.rst
```diff
@@ -703,3 +703,30 @@ functions but for round specific needs, such as ``start_of_round``,
...
+
+The replay file can be read by the ``stechec2-replay`` utility, which only
+requires the rules shared library to be present to replay a match::
+
+   stechec2-replay --replay match.replay --rules rules.so
+
+The replay file contains the map, if the game uses one, the list of actions
+sent by the clients (champions and spectators) and the final score.  At the end
+of the replayed match, ``stechec2-replay`` checks if the obtained score matches
+the score save in the reply file.
```
zopieux: saved

doc/rules.rst
```diff
@@ -703,3 +703,30 @@ functions but for round specific needs, such as ``start_of_round``,
...
+
+You can use this tool to replay a match that triggers a specific bug, or use
+the replay file a unit test.
```
zopieux: I think you a word or two

src/lib/rules/player.hh
```diff
@@ -60,6 +61,19 @@ struct Players final : utils::IBufferizable
         }
     }
 
+    std::string scores_yaml() const
+    {
+        std::stringstream ss;
+        for (auto& player : players)
+        {
+            ss << "---" << std::endl
+               << "player: " << player->name.c_str() << std::endl
```
zopieux, line 71: how nasty (how user controlled) can this name be? does it need escaping and/or quoting at least?

zopieux (commented):
I'll have to actually toy with this locally before stamping but otherwise looks good.

src/lib/rules/player.hh
audebert: It's a free string (incl. \0 I think). A rogue client could potentially send a specially crafted name to overwrite their score here.

Zopieux: Yeah or naming their champion 123 without bad intentions, which converts to an int, which is not what we want.

src/lib/rules/replay-messenger.hh
```diff
@@ -0,0 +1,25 @@
...
+#include <rules/messenger.hh>
+
+namespace rules {
+
+class Actions;
+
+class ReplayMessenger
+{
+public:
+    ReplayMessenger(utils::Buffer* buf) : buf_(buf) {}
```
zopieux: explicit

src/lib/rules/replay-messenger.cc
```diff
@@ -0,0 +1,15 @@
+#include "replay-messenger.hh"
+
+#include "rules/actions.hh"
+
+namespace rules {
```
zopieux, line 5: Nit: newlines around

src/lib/rules/rules.cc
```diff
@@ -276,6 +307,53 @@ void TurnBasedRules::player_loop(ClientMessenger_sptr msgr)
...
 
+void TurnBasedRules::replay_loop(ReplayMessenger_sptr msgr)
+{
+    at_start();
+
+    auto actions = get_actions();
+    while (!is_finished())
+    {
+        start_of_round();
+        for (const auto& p : players_->players)
```
zopieux: Nit: I prefer slightly longer variable names, what about `player`?

src/lib/rules/rules.cc
```diff
@@ -276,6 +307,53 @@ void TurnBasedRules::player_loop(ClientMessenger_sptr msgr)
...
+            msgr->pull_actions(actions);
+            DEBUG("Pulled %d actions", actions->size());
+            for (const auto action : actions->actions())
+                apply_action(action);
+            actions->clear();
+
+            end_of_player_turn(p->id);
+            end_of_turn(p->id);
+
+            for (const auto& s : spectators_->players)
```
zopieux: Same here

wscript
```diff
@@ -237,3 +239,15 @@ def build_server(bld):
...
                 use=['stechec2', 'gflags'])
+
+def build_replay(bld):
+    bld.program(source='''
+            src/replay/main.cc
+            src/replay/replay.cc
+        ''',
+                target='stechec2-replay',
+                defines=[
+                    'MODULE_COLOR=ANSI_COL_RED', 'MODULE_NAME="replay"',
```
zopieux, line 250: Another color maybe? This is already used by the server.

audebert: The idea behind stechec2-replay is that it is exactly like stechec2-server, just without clients.

It's just implemented as a separate bin here for simplicity, but another implem could have been to re-use the same stechec2-server and feed the replayed actions with a new "stechec2-replay-client" instead.

wscript
```diff
@@ -236,3 +238,15 @@ def build_server(bld):
                     'MODULE_VERSION="%s"' % VERSION
                 ],
                 use=['stechec2', 'gflags'])
+
```
zopieux, line 241: Missing newline.

zopieux (approved):

PR #66, opened 2020-01-22
Jinja draft for reviews

Shaac:


tools/generator/make_rules.py
```diff
@@ -0,0 +1,53 @@
...
+    if value.endswith(" array"):
+        return "std::vector<{}>".format(cxx_type(value[:-6]))
+    return value
+
+def cxx_args(value) -> str:
+    return ", ".join("{} {}".format(cxx_type(i[1]), i[0]) for i in value)
+
+def snake_to_camel_case(value: str) -> str:
+    """Convert a snake case identifier to a upper camel case one"""
+    return "".join(i.capitalize() for i in value.split("_"))
```
seirl, line 23: I think it would be better to have a |camel_case and a |snake_case filter that are both idempotent and accept both cases as their input. It sounds doable.

tools/generator/make_rules.py
```diff
@@ -0,0 +1,53 @@
...
+
+    for copy in ["entry.cc", "game_state.cc", "game_state.hh"]:
+        copyfile(os.path.join(template_folder, copy), os.path.join(install_path, copy))
+
+    for gen in ["actions.hh", "api.cc", "api.hh", "constant.hh", "interface.cc", "rules.hh", "rules.cc"]:
+        env.get_template(gen + ".jinja2").stream(yaml=conf).dump(os.path.join(install_path, gen))
+
+    for action in conf["function"]:
+        if "fct_action" in action and action["fct_action"]:
+            env.get_template("action_template.hh.jinja2").stream(func=action).dump(os.path.join(install_path, "action_") + action["fct_name"] + ".hh")
```
seirl: I'd call the template with `action=action`, it's a lot clearer in the templates

tools/generator/files/action_template.hh.jinja2
```diff
@@ -0,0 +1,42 @@
...
+#include "constant.hh"
+#include "game_state.hh"
+
+{% set args = func.fct_arg + [['player_id', 'int', '']] -%}
+class Action{{ func.fct_name|snake_to_camel_case }} : public rules::Action<GameState>
+{
+public:
+    {%- set comma = joiner("    , ") %}
+    Action{{ func.fct_name|snake_to_camel_case }}({{ args|cxx_args }})
+        : {% for arg in args %}{{ comma() }}{{ arg[0] }}_({{ arg[0] }})
```
seirl: Use unpacking here?

`for arg_name, _, _ in args`

tools/generator/files/action_template.hh.jinja2
```diff
@@ -0,0 +1,42 @@
...
+    }
+    Action{{ func.fct_name|snake_to_camel_case }}() {} // for register_action()
+
+    int check(const GameState& st) const override;
+    void apply_on(GameState* st) const override;
+
+    void handle_buffer(utils::Buffer& buf) override
+    {
+        {%- for arg in args %}
+        buf.handle({{ arg[0] }}_);
```
seirl: ditto, unpacking

tools/generator/files/action_template.hh.jinja2
```diff
@@ -0,0 +1,42 @@
...
+        {%- endfor %}
+    }
+
+    uint32_t player_id() const override { return player_id_; };
+    uint32_t id() const override { return ID_ACTION_{{ func.fct_name|upper }}; }
+
+private:
+    {%- for arg in args %}
+    {{ arg[1] }} {{ arg[0] }}_;
```
seirl: ditto, `for arg_name, arg_type in args`

tools/generator/make_rules.py
```diff
@@ -0,0 +1,53 @@
...
+    env = jinja2.Environment(loader=jinja2.FileSystemLoader(searchpath=template_folder))
+    env.filters["cxx_type"] = cxx_type
+    env.filters["cxx_args"] = cxx_args
+    env.filters["snake_to_camel_case"] = snake_to_camel_case
+
+    for copy in ["entry.cc", "game_state.cc", "game_state.hh"]:
+        copyfile(os.path.join(template_folder, copy), os.path.join(install_path, copy))
+
+    for gen in ["actions.hh", "api.cc", "api.hh", "constant.hh", "interface.cc", "rules.hh", "rules.cc"]:
+        env.get_template(gen + ".jinja2").stream(yaml=conf).dump(os.path.join(install_path, gen))
```
seirl: i'd say `game=conf`, `yaml` is a bit confusing in the templates

tools/generator/files/api.hh.jinja2
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2012-2020 Association Prologin <association@prologin.org>
+
+#pragma once
+
+#include <memory>
+{#- The following two headers may not be necessary. They are only used if there
+is a std::string, or a std::vector in the generated function signatures.
+We could make the test here and add the includes only if necessary, but it does
+not cost must to include them in every cases. #}
```
seirl: I think you can remove this comment, because it's probably going to show up in a lot of files, and it doesn't really matter as you say.

tools/generator/files/api.hh.jinja2
```diff
@@ -0,0 +1,46 @@
...
+class Api final : public rules::Api<GameState, error>
+{
+public:
+    Api(std::unique_ptr<GameState> game_state, rules::Player_sptr player);
+    ~Api() {}
+
+{#- Generate the C++ declaration for each API function -#}
+{% for func in yaml.function %}
+{# Since this is a header, generate the function documentation #}
+    {{ ("/// " + func.fct_summary)|wordwrap(79, False, "\n    /// ") }}
```
seirl: Maybe this could deserve its own filter? `func|cxx_comment` with a `doc=False` argument for `///`?

tools/generator/files/constant.hh.jinja2
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2012-2020 Association Prologin <association@prologin.org>
+
+#pragma once
+
+#include <functional> // needed for std::hash
+
+{%- for constant in yaml.constant %}
+{{ ("/// " + constant.cst_comment)|wordwrap(79, False, '\n/// ') }}
```
seirl: See, other uses of the filter :-P

tools/generator/files/constant.hh.jinja2
```diff
@@ -0,0 +1,33 @@
...
+#define {{ constant.cst_name }} {{ constant.cst_val }}
+{% endfor %}
+
+{%- for enum in yaml.enum %}
+
+{{ ("/// " + enum.enum_summary)|wordwrap(79, False, '\n/// ') }}
+typedef enum {{ enum.enum_name }}
+{
+    {%- set padding = "%%-%ds"|format(enum.enum_field|map('first')|map('length')|max + 1) %}
+    {%- for field in enum.enum_field %}
```
seirl: Unpacking again

tools/generator/files/constant.hh.jinja2
```diff
@@ -0,0 +1,33 @@
...
+    {{ padding | format(field[0]|upper + ",") }} ///< {{ field[1] }}
+    {%- endfor %}
+} {{ enum.enum_name }};
+{% endfor %}
+
+{%- for struct in yaml.struct %}
+{{ ("/// " + struct.str_summary)|wordwrap(79, False, '\n/// ') }}
+typedef struct {{ struct.str_name }}
+{
+    {%- for field in struct.str_field %}
```
seirl: Unpacking

seirl (changes requested):

tools/generator/make_rules.py
Shaac: I don’t really see the point.
The functions are written in the yaml in snake_case. And in some places, it will be converted in camel case. That is why I named the function `snake_to_camel_case` instead of the previous `camel_case`.
So the name is a bit long, but it is the only case transformation we’ll be needing.

tools/generator/make_rules.py
Shaac: ok. I wanted to keep the same variable name as in other templates (when I do ` for func in yaml.function`) but maybe it is not important.

tools/generator/files/api.hh.jinja2
Shaac: ok, I added this in the very beginning but I’ve been less verbose since

tools/generator/files/api.hh.jinja2
Shaac: I’m not sure about this. Here it is really a job for the template engine, since it is just doing a wordwrap.

Also, I did not understood the `doc=False`

tools/generator/files/api.hh.jinja2
```diff
@@ -0,0 +1,42 @@
...
+{% for func in game.function %}
+{# Since this is a header, generate the function documentation #}
+    {{ ("/// " + func.fct_summary)|wordwrap(79, False, "\n    /// ") }}
+
+{%- if func.fct_action %}
+{#- Action functions are defined in their own cpp. Here we just use a wrapper
+    to bind them to the API #}
+    ApiActionFunc<Action{{ func.fct_name|camel_case }}> {{ func.fct_name }}{this};
+{%- else %}
+{#- Other action are just declared here, their definition will be in api.cc #}
```
seirl: Other API methods*, "action" is confusing here

tools/generator/files/api.hh.jinja2
```diff
@@ -0,0 +1,42 @@
...
+    ~Api() {}
+
+{#- Generate the C++ declaration for each API function -#}
+{% for func in game.function %}
+{# Since this is a header, generate the function documentation #}
+    {{ ("/// " + func.fct_summary)|wordwrap(79, False, "\n    /// ") }}
+
+{%- if func.fct_action %}
+{#- Action functions are defined in their own cpp. Here we just use a wrapper
+    to bind them to the API #}
```
seirl, line 34: Have you considered having these as proper C++ comments? I think they might be useful for people reading the generated code, not only people working on the templates. This is valid for most of the other comments I think.

tools/generator/files/constant.hh.jinja2
```diff
@@ -0,0 +1,33 @@
...
+{% endfor %}
+
+{%- for enum in game.enum %}
+
+{{ ("/// " + enum.enum_summary)|wordwrap(79, False, '\n/// ') }}
+typedef enum {{ enum.enum_name }}
+{
+    {%- set padding = "%%-%ds"|format(enum.enum_field|map('first')|map('length')|max + 1) %}
+    {%- for field_name, field_comment in enum.enum_field %}
+    {{ padding | format(field_name|upper + ",") }} ///< {{ field_comment }}
```
seirl: Just remove the padding imo

tools/generator/files/interface.cc.jinja2
```diff
@@ -0,0 +1,76 @@
...
+    api->{{ func.fct_name }}({{ func.fct_arg|map('first')|join(', ') }});
+}
+{%- endfor %}
+
+{%- for enum in game.enum %}
+std::ostream& operator<<(std::ostream& os, {{ enum.enum_name }} v)
+{
+    switch (v)
+    {
+    {%- for field in enum.enum_field %}
```
seirl: Unpacking

tools/generator/files/interface.cc.jinja2
```diff
@@ -0,0 +1,76 @@
...
+
+{%- for struct in game.struct %}
+std::ostream& operator<<(std::ostream& os, {{ struct.str_name }} v)
+{
+    os << "{ ";
+    {%- for field in struct.str_field %}
+    {%- if not loop.first %}
+    os << ", ";
+    {%- endif %}
+    os << "{{ field[0] }}"
```
seirl: Unpacking

tools/generator/make_rules.py
```diff
@@ -0,0 +1,53 @@
...
+
+def cxx_type(value: str) -> str:
+    if value == "string":
+        return "std::string"
+    if value.endswith(" array"):
+        return "std::vector<{}>".format(cxx_type(value[:-6]))
+    return value
+
+def cxx_args(value) -> str:
+    return ", ".join("{} {}".format(cxx_type(i[1]), i[0]) for i in value)
```
seirl: Unpacking

zopieux:
Random idea: what about giving zero damn to the style (indents, whitespaces) inside the templates and instead pipe the output inside clang-format? This is only ever gonna output C++ right?

* no controversial debates over style, just choose a sensibly .clang-format and stick with it
* no need for jinja boilerplate for indenting properly; jinja was designed to output HTML which has virtually no significant whitespacing
* no need to waste time fixing incorrect whitespacing that will, without the shadow of a doubt, happen regularly
* no need to care about flowing paragraphs in  comment!

seirl (changes requested):

tools/generator/files/api.hh.jinja2
Shaac: That could be an idea, but they’ll have to be different. Here we’re in the middle of a template loop.

seirl (approved):

PR #68, opened 2020-01-22
Refactor Players_sptr

audebert:
Players_sptr was std::shared_ptr<std::vector<std::shared_ptr<Player>>>.
There was no reason to have a shared_ptr holding other shared_ptrs.

This commit removes the Players_sptr type.

Also, this commit replaces the Player_sptr typedef with the underlying
type: std::shared_ptr<Player>. Rationale: let's be explicit about the
types.

Also, move the Players field of the game's GameState to the base class
rules::GameState.

### DIFF
--- a/src/lib/rules/player.hh
+++ b/src/lib/rules/player.hh
@@ -2,18 +2,20 @@
 // Copyright (c) 2012 Association Prologin <association@prologin.org>
 #pragma once
 
+#include <algorithm>
 #include <cstdint>
 #include <memory>
 #include <sstream>
 #include <string>
-#include <utils/buffer.hh>
 #include <utility>
+#include <utils/buffer.hh>
 #include <vector>
 
 namespace rules {
 
-struct Player
+struct Player final
 {
+    Player() = default;
     Player(uint32_t id_, uint32_t type_)
         : id(id_), type(type_), score(0), nb_timeout(0), name("anonymous")
     {}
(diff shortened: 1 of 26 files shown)


src/lib/rules/player.hh
```diff
@@ -33,38 +35,40 @@ struct Player
...
-
-struct Players final : utils::IBufferizable
+class Players final : public utils::IBufferizable
 {
+    using PlayersVector = std::vector<std::shared_ptr<Player>>;
+
+public:
     Players() = default;
-    Players(std::vector<Player_sptr> players) : players(players) {}
+    Players(PlayersVector players) : players_{players} {}
```
zopieux: If you're taking a copy like now, std::move() it? Or take a const ref so that {} is the copy operator.

src/lib/rules/player.hh
```diff
@@ -33,38 +35,40 @@ struct Player
...
 
     void handle_buffer(utils::Buffer& buf) override
     {
         if (buf.serialize())
         {
-            for (auto& p : players)
+            for (const auto& p : players_)
```
zopieux: The usual naming nit, but maybe this was already done in master.

src/lib/rules/player.hh
```diff
@@ -33,38 +35,40 @@ struct Player
...
             {
                 // Get a player
-                Player_sptr player = Player_sptr(new Player(0, 0));
+                auto player = std::make_shared<Player>();
 
                 // And unserialize it
                 player->handle_buffer(buf);
-                players.push_back(player);
+
+                players_.push_back(player);
```
zopieux, line 67: Kinda orthogonal to this change but this pushes back the players without clearing players_ first, is this intended?

src/lib/rules/player.hh
```diff
@@ -74,12 +78,22 @@ struct Players final : utils::IBufferizable
...
+    void sort()
+    {
+        std::sort(players_.begin(), players_.end(),
+                  [](const auto& a, const auto& b) { return a->id < b->id; });
+    }
+
+    size_t size() const { return players_.size(); }
 
-using Players_sptr = std::shared_ptr<Players>;
+    const PlayersVector& all() const { return players_; }
```
zopieux: Optional: implement the iterator "trait" instead? This looks like this method is here to hide the fact this class is just a proxy for `vector<Player>`.

zopieux (commented):
Looks nice except for that nasty `all()` that looks like you're being lazy implementing more STL container methods :-)

src/lib/rules/player.hh
audebert: Fixed.

src/lib/rules/player.hh
audebert: Added clear().

audebert (commented):
Removed the all() calls.

PR #69, opened 2020-01-22
lib, server: fix zmq send()/recv() deprecation

zopieux:


zopieux:
On hold: this is way too modern for our current dependency pinning on libzmq3.

juli0z:
Migrated to https://gitlab.com/prologin/tech/tools/stechec2/-/merge_requests/69

PR #70, opened 2020-01-22
lib: apply many clang-tidy suggestions; fix typos

zopieux:


wscript
```diff
@@ -180,7 +180,6 @@ def build_lib(bld):
             src/lib/net/message.cc
 
             src/lib/rules/actions.cc
-            src/lib/rules/messenger.cc
```
audebert, line 183: Let's remove that file altogether: https://github.com/prologin/stechec2/blob/master/src/lib/rules/messenger.cc

audebert: Disregard that comment -- this PR already deletes it.

audebert (approved):

PR #71, opened 2020-01-23
Use Python generator by default, use Ruby as fallback

seirl:


tools/generator/rules.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def cxx_args(value) -> str:
+    return ", ".join("{} {}".format(cxx_type(type_), name)
+                     for [name, type_, _] in value)
+
+
+def camel_case(value: str) -> str:
+    """Convert a snake case identifier to a upper camel case one"""
+    return "".join(i.capitalize() for i in value.split("_"))
```
zopieux, line 20: Nit: move outside the cxx_ stuff.

tools/generator/rules.py
```diff
@@ -0,0 +1,58 @@
...
+                "interface.cc", "rules.hh", "rules.cc", "entry.cc",
+                "game_state.cc", "game_state.hh"]:
+        env.get_template(gen + ".jinja2").stream(game=game).dump(os.path.join(
+            out_dir, gen))
+
+    for action in game["function"]:
+        if "fct_action" in action and action["fct_action"]:
+            (env.get_template("action_template.hh.jinja2")
+             .stream(action=action)
+             .dump(os.path.join(out_dir, "action_")
```
zopieux, line 54: Please refactor this by splitting the dir and filename in dedicated variables. It's not very readable. Same below.

Shaac:
We need to add `python-jinja` as dependency.
Both in the doc, and in the PKGBUILD.

Shaac (changes requested):
We need to add `python-jinja` as a dependency.
Both in the doc, and in the PKGBUILD.

PR #73, opened 2020-01-23
generator: automatically register filters

seirl:
Could you please comment on whether you think the amount of boilerplate is reasonable?

There are a lot of redundancies when using the generic filters, but I'm not sure how we could do that better.

### DIFF
--- a/tools/generator/filters/c.py
+++ b/tools/generator/filters/c.py
@@ -1,7 +1,9 @@
-import functools
-from .common import generic_args, generic_prototype
+import jinja2
+from . import register_filter
+from .common import generic_args, generic_prototype, generic_comment
 
 
+@register_filter
 def c_type(value: str) -> str:
     if value == "string":
         return "char*"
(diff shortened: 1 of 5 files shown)


tools/generator/filters/c.py
```diff
@@ -10,5 +12,17 @@ def c_type(value: str) -> str:
...
+
+@register_filter
+def c_prototype(value):
+    return generic_prototype(value, arg_mapper=c_args)
+
+
+@register_filter
+@jinja2.environmentfilter
+def c_comment(env, value, indent: int = 0):
+    return generic_comment(env, value=value, start='// ', indent=indent)
```
seirl, line 28: This whole section is the boilerplate I'm talking about, that will basically need to be copypasted across all languages.

Yes, I know how to make it more generic, but it comes at the cost of readability. We know how unreadable it can become in the case of the previous generators when too much magic is involved.

zopieux: I don't mind verbosity when it's making stuff more maintainable in the long run.

seirl: Alternative "magic" option to replace all these lines:

```
c_args = register_partial_filter('c_args', generic_args, type_mapper=c_type)
c_args = register_partial_filter('c_prototype', generic_args, type_mapper=c_args)
c_comment = register_partial_filter('c_comment', generic_comment, start='// ')
```

Shaac (approved):
The minimum boilerplate seems good.
I’m not a big fan of big centralizing global variable like `FILTER_LIBRARY`, but I guess it’s OK in our use case.

We’ll just have to keep in mind we have to update `load_library` at each new filter file.

Issue #48, opened 2019-06-06
Generate the wscript for the rules


Shaac:
Writing the `wscript` for a new game is always the same. Let’s generate it automatically (maybe with `stechec2-generator rules`).

For reference, in the docs:
https://stechec2.readthedocs.io/en/latest/rules.html#the-wscript
There is also often a section for the unit tests, but it is not in the docs!

PR #74, opened 2020-01-31
Generate wscript for the rules

Shaac:
Implement #48


tools/generator/templates/rules/wscript.jinja2
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python3
```
seirl, line 1: I don't think we need a shebang actually, it's not supposed to be a main

tools/generator/rules.py
```diff
@@ -21,14 +25,16 @@ def make_rules(game, out_dir: Path) -> None:
...
-                     out_name='action_{}.hh'.format(action['fct_name']),
-                     action=action)
-        gen.template('action_template.cc',
-                     out_name='action_{}.cc'.format(action['fct_name']),
-                     action=action)
+        for ext in ('hh', 'cc'):
+            gen.template('action_template.' + ext,
+                         out_name=os.path.join(
+                             'src',
+                             'action_{}.{}'.format(action['fct_name'], ext)),
```
seirl: I think you can just use `src/action_{}.{}` here instead of the `join()`.

seirl (changes requested):

tools/generator/templates/rules/wscript.jinja2
```diff
@@ -0,0 +1,58 @@
...
+        defines = ['MODULE_COLOR=ANSI_COL_BROWN', 'MODULE_NAME="rules"'],
+        target = '{{ game.name }}',
+        use = ['stechec2'],
+    )
+
+    abs_pattern = os.path.join(bld.path.abspath(), 'src/tests/test-*.cc')
+    for test_src in glob.glob(abs_pattern):
+
+        test_name = os.path.split(test_src)[-1]
+        test_name = test_name[5:-3]
```
zopieux: This looks brittle. Can you make it explicit what these lengths are, eg. by using len("the thing")?

tools/generator/templates/rules/wscript.jinja2
Shaac: You are right we don’t need it. It kept it because we use it everywhere.

It does have a use though: it allows editors to know that this file is written in python. Without it there is no syntax highlighting and such. So it might be a good idea to use it.

tools/generator/rules.py
Shaac: ok

tools/generator/templates/rules/wscript.jinja2
Shaac: done

tools/generator/rules.py
```diff
@@ -21,14 +25,15 @@ def make_rules(game, out_dir: Path) -> None:
...
         if not action.get("fct_action"):
             continue
-        gen.template('action_template.hh',
-                     out_name='action_{}.hh'.format(action['fct_name']),
-                     action=action)
-        gen.template('action_template.cc',
-                     out_name='action_{}.cc'.format(action['fct_name']),
-                     action=action)
+        for ext in ('hh', 'cc'):
+            gen.template(os.path.join('src', 'action_template.' + ext),
```
seirl: nit: `'src/action_template.{}'.format(ext)` for consistency

tools/generator/rules.py
```diff
@@ -21,14 +25,15 @@ def make_rules(game, out_dir: Path) -> None:
         'game_state.hh'
     ]
     for tpl in base_files:
-        gen.template(tpl)
+        gen.template(os.path.join('src', tpl))
```
seirl: nit: I'd rather have the list of `base_files` be prefixed by `src/` and include `wscript` so you don't have to special case it and it's more explicit.

tools/generator/generator.py
```diff
@@ -23,7 +26,9 @@ def template(self, name, out_name=None, **params):
             out_name = name
         tpl = self.env.get_template(name + '.jinja2')
         out = tpl.stream(game=self.game, **params)
-        out.dump((self.out_dir / out_name).open('w'))
+        out_path = self.out_dir / out_name
+        out_path.parent.mkdir(parents=True, exist_ok=True)
+        out.dump(out_path.open('w'))
```
seirl, line 31: (I just realized there's a fd leaking here, although it's not because of this PR)

seirl (approved):
Approved with a few minor quibbles, feel free to merge after fixing.

tools/generator/rules.py
Shaac: done

tools/generator/rules.py
Shaac: done

PR #76, opened 2020-02-13
Add C generator

seirl:


tools/generator/templates/player/c/Makefile.jinja2
```diff
@@ -0,0 +1,44 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
```
seirl, line 1: Almost nothing is done in this file yet, because we need to figure out how to do it properly first

tools/generator/filters/common.py
```diff
@@ -1,7 +1,10 @@
 import textwrap
 from functools import partial
+from jinja2 import contextfunction
 
-from . import register_filter, register_test
+from . import register_filter, register_test, register_function
+
+SIMPLE_TYPES = ('int', 'float', 'string')
```
seirl: Needs to be 'double'

tools/generator/templates/player/c/interface.cc.jinja2
```diff
@@ -0,0 +1,125 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2020 Association Prologin <association@prologin.org>
+
+// This file is generated. It contains the code to call the API functions from
+// the C++ language. Do not edit this file.
```
Shaac: for the C language

tools/generator/templates/player/c/interface.cc.jinja2
```diff
@@ -0,0 +1,125 @@
...
+
+#include <vector>
+#include <string>
+
+extern "C" {
+#include "api.h"
+}
+
+template<typename Lang, typename Cxx>
+Cxx lang2cxx(Lang in)
```
Shaac: I don’t really like the name of this function. I find using `lang` instead of `c` harder to read. Same for `Lang`.

Also it would be nice to run clang-format on the template.

tools/generator/templates/player/c/Makefile.jinja2
Shaac: Don’t we want to at least add the C sources into the objs variable before merging?

Shaac (approved):

Shaac:
Also don’t forget to add the templates to be installed in the wscript.

tools/generator/templates/player/c/api.h.jinja2
```diff
@@ -0,0 +1,51 @@
...
+{% endfor %}
+
+{# Forward declaration of structs to build arrays #}
+{% for struct in game.struct %}
+struct {{ struct.str_name }};
+{% endfor %}
+{% for type in array_types(game) %}
+
+typedef struct {{ type }}_array {
+    {%+ if is_struct(type) %}struct {% endif %}{{ type|c_type }}* datas;
```
Shaac: shouldn’t data be singular?

tools/generator/filters/common.py
```diff
@@ -41,3 +59,8 @@ def generic_comment(value: str, start: str, indent: int = 0) -> str:
         break_long_words=False,
         replace_whitespace=False
     ))
+
+
+@register_function
+def array_types(game):
+    return SIMPLE_TYPES + tuple([s['str_name'] for s in game['struct']])
```
Shaac: Maybe we should filter out the array types not used in the game?

For instance, for tictactoe we declare double_array, but there is no double.

tools/generator/templates/player/c/interface.cc.jinja2
```diff
@@ -0,0 +1,125 @@
...
+#include <string>
+
+extern "C" {
+#include "api.h"
+}
+
+template<typename CType, typename CxxType>
+CxxType c_to_cxx(CType in)
+{
+  return in;
```
Shaac: From here to line 59, there should be indentation of 4 spaces.

tools/generator/templates/player/c/interface.cc.jinja2
```diff
@@ -0,0 +1,125 @@
...
+{% for func in game.function %}
+
+extern "C" {{ func|c_internal_cxx_prototype(prefix='api_') }};
+
+extern "C" {{ func|c_prototype }}
+{
+    {% if func is returning %}
+    return {{ func.fct_ret_type|cxx_to_c }}({% endif -%}
+    api_{{ func.fct_name }}(
+        {% for arg_name, arg_type, arg_comment in func.fct_arg %}
```
Shaac: Maybe `{%-`, the newline looks very strange in the generated code

tools/generator/templates/player/c/interface.cc.jinja2
```diff
@@ -0,0 +1,125 @@
...
+        {{ arg_type|c_to_cxx }}({{ arg_name }}){% if not loop.last %},{% endif %}
+        {% endfor %}
```
Shaac: Maybe `-%}` as well

tools/generator/templates/player/c/Makefile.jinja2
```diff
@@ -0,0 +1,40 @@
...
+# Copyright (c) 2020 Association Prologin <association@prologin.org>
+
+# Add all your sources (must be suffixed by *.cc) in here, separated by spaces
```
Shaac: suffixed by *.c

tools/generator/templates/player/c/Makefile.jinja2
```diff
@@ -0,0 +1,40 @@
...
+CHAMPION_HEADERS =
+
+# ----
+# Do not modify the lines below
+
+INTERFACE_SOURCES = interface.cc
+INTERFACE_HEADERS = api.hh
+
+CPPFLAGS = -MMD -MP
+CXXFLAGS = -ggdb3 -Wall -std=c++17 -fPIC
```
Shaac: We should add similar C flags.

In old generators the final command was `gcc -ggdb3 -Wall -Wextra -std=c11 -MD -MP -MF ./.prologin.o.d  -fPIC -c prologin.c -o prologin.o`

tools/generator/templates/player/c/api.h.jinja2
seirl: This is copied from the old generators but yes. I think "items" is even better.

tools/generator/filters/common.py
seirl: Yes, probably a good idea.

PR #77, opened 2020-02-16
rubygenerator: remove toposort

seirl:


Shaac (approved):
I’m not expert, but this seems good. And removing this dirty file is a good thing.

PR #78, opened 2020-02-16
rubygenerator: OCaml: closures are now const

seirl:


Shaac:
It fixes compilation issues for me.
When did this broke?

seirl:
https://github.com/ocaml/ocaml/pull/2293

> This change is compatible with all versions of OCaml: it is always safe to store a `value*` in a variable of type `const value*`. So, it is not necessary to maintain two versions of C stubs to cleanly compile both before and after this patch, but C stubs will generate a warning until they're updated.

PR #81, opened 2020-02-18
Apply common style.

### LINKED ISSUE #79: Use common clang-format
Replace current clang-format configuration with https://github.com/prologin/prologin2019/blob/master/.clang-format and reformat all the codebase.

audebert:
Fixes #79

### DIFF
--- a/games/plusminus/champions/cxx/prologin.hh
+++ b/games/plusminus/champions/cxx/prologin.hh
@@ -20,7 +20,8 @@ typedef enum error
 // This is needed for old compilers
 namespace std
 {
-template <> struct hash<error>
+template <>
+struct hash<error>
 {
     size_t operator()(const error& v) const
     {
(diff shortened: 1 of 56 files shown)


games/plusminus/champions/cxx/prologin.hh
```diff
@@ -82,15 +82,16 @@ struct hash<std::vector<T>>
...
 } // namespace std
 
-extern "C" {
+extern "C"
+{
 
-/// Function called at the start of the game
-void init_game();
+    /// Function called at the start of the game
+    void init_game();
```
Shaac, line 91: Maybe we want to add an option to not indent extern C. The same way we don’t indent namespaces.

tools/rubygenerator/files/rules.cc
```diff
@@ -8,7 +8,7 @@
 #include "actions.hh"
 
 Rules::Rules(const rules::Options opt)
-    :!!rules_cc_inheritance!! sandbox_(opt.time)
+    : !!rules_cc_inheritance !!sandbox_(opt.time)
```
Shaac: This is a ~template. This will probably break the ruby generator.

audebert: Nice catch!

games/plusminus/champions/cxx/prologin.hh
audebert: Done.

PR #82, opened 2020-02-18
rules: client id refactor

### LINKED ISSUE #67: Setup player IDs to be natural array indexes
For a 2 player game, ensure that the first player id is 0, second player id is 1. This will avoid having to implement more get_player_id-type helpers (https://github.com/prologin/prologin2019/blob/091e27c4c172a56ac5d6d3f75f3cf17547fc3de8/src/game_state.cc#L49).

This is a non-trivial change as the server currently allocates player and specator IDs first-come first served and does not know how many player games are for.

audebert:
Fixes #67 

This PR introduces a new exported function: rules_config(). It enable the game to specify the number of expected players. With that information, the server can allocate the player client_ids to be in natural order: the first player will have id 0, second 1, etc. This will introduce the guarantee that player ids can be used as vector indices (0-indexed and contiguous), and therefore simplify games logic.

### DIFF
--- a/doc/rules.rst
+++ b/doc/rules.rst
@@ -39,6 +39,7 @@ First, we must write some config boilerplate at the top of the file::
 
   name: connect4  # The name of the game
   rules_type: turnbased # The type of rules to follow
+  player_count: 2 # The number of players
 
   constant:
     # Place your constants here
--- a/doc/rules.rst
+++ b/doc/rules.rst
@@ -38,8 +38,8 @@ Write the YAML
 First, we must write some config boilerplate at the top of the file::
 
   name: connect4  # The name of the game
-  rules_type: turnbased # The type of rules to follow
-  player_count: 2 # The number of players
+  rules_type: turnbased  # The type of rules to follow
+  player_count: 2  # The number of players
 
   constant:
     # Place your constants here
--- a/games/tictactoe/src/tests/test-rules.cc
+++ b/games/tictactoe/src/tests/test-rules.cc
@@ -36,10 +36,10 @@ TEST_F(RulesTest, Rules_FinishDraw)
         rules->game_state().set_cell(to_play1[idx1++], PLAYER_1);
         rules->end_of_player_turn(PLAYER_1);
 
-        if (idx1 == to_play1.size())
-            EXPECT_TRUE(rules->is_finished());
-        else
-            EXPECT_FALSE(rules->is_finished());
+        // The game is finished only after player 2 has played. There is not
+        // support for finishing a game in the middle of a round in the server
+        // loop.
+        EXPECT_FALSE(rules->is_finished());
 
         rules->start_of_player_turn(PLAYER_2);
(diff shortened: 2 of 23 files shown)


games/tictactoe/src/tests/test-rules.cc
```diff
@@ -36,10 +36,10 @@ TEST_F(RulesTest, Rules_FinishDraw)
         rules->game_state().set_cell(to_play1[idx1++], PLAYER_1);
         rules->end_of_player_turn(PLAYER_1);
 
-        if (idx1 == to_play1.size())
-            EXPECT_TRUE(rules->is_finished());
-        else
-            EXPECT_FALSE(rules->is_finished());
+        // The game is finished only after player 2 has played. There is not
```
zopieux: s/not/no/

src/server/server.cc
```diff
@@ -119,13 +126,23 @@ bool used_identifier(uint32_t player_id, const rules::Players& players)
...
 
-void Server::wait_for_players()
+void Server::wait_for_clients()
 {
     if (FLAGS_nb_clients <= 0)
         FATAL("Server started with nb_clients <= 0.");
 
+    int spectator_count = FLAGS_nb_clients - config_.player_count;
+    if (spectator_count < 0)
+        FATAL("Server must be started with at least -nb_clients %d",
```
zopieux: Suggestion: use `--foo` for flags instead of `-foo`, it looks like a minus (and it's not GNU).

src/server/server.cc
```diff
@@ -39,9 +40,15 @@ void Server::run()
 
     INFO("Server Started");
 
+    // The rules configure the server
+    rules_config(&config_);
+    NOTICE("Game configuration:");
+    NOTICE("- name: %s", config_.name);
+    NOTICE("- %d players", config_.player_count);
```
zopieux, line 47: Would it be annoying to also rename all the *player* variables that are actually representing *clients* into, well, *clients*? `client_count`, `ClientType`, `client_type`, `client_id`, etc.

src/server/server.cc
audebert: Done.

src/server/server.cc
audebert: Yes we should. I've added a TODO to refactor rules::Player into rules::Client and will address that in another PR.

games/tictactoe/src/tests/test-rules.cc
audebert: Done.

doc/rules.rst
```diff
@@ -39,6 +39,7 @@ First, we must write some config boilerplate at the top of the file::
 
   name: connect4  # The name of the game
   rules_type: turnbased # The type of rules to follow
+  player_count: 2 # The number of players
```
zopieux: Nit: two spaces before # everywhere

audebert: Done

PR #84, opened 2020-02-22
Add PHP generator

seirl:


tools/generator/filters/__init__.py
```diff
@@ -31,7 +31,8 @@ def load_library_in(env):
         common,
         c,
         cxx,
-        python
+        python,
+        php
```
Shaac: Isn’t it better alphabetically?

.github/workflows/stechec2-generator-test.yml
```diff
@@ -14,11 +14,15 @@ jobs:
...
+        # Some packages here are included in the --ldflags of php-config but
+        # not present by default in the CI, so we install them manually.
+        # (libargon2-1 and libsodium-dev)
         run: |
           sudo apt-get update
           sudo apt-get install \
             python3 python3-dev python3-yaml python3-jinja2 flake8 php-cli \
-            php-dev libphp-embed ocaml mono-devel ghc openjdk-8-jdk
+            php-dev libphp-embed ocaml mono-devel ghc openjdk-8-jdk \
+            libargon2-0-dev libsodium-dev  # XXX: to remove
```
Shaac: Why to remove if we need them for php-config?

tools/generator/player.py
```diff
@@ -24,3 +24,5 @@ def make_player(game, out_dir: Path) -> None:
              ['api.h', 'champion.c', 'interface.cc', 'Makefile'])
     gen_lang(game, out_dir, 'python',
              ['api.py', 'Champion.py', 'interface.cc', 'Makefile'])
+    gen_lang(game, out_dir, 'php',
```
Shaac, line 25: Same, I’d rather have them alphabetically.

Shaac (approved):

.github/workflows/stechec2-generator-test.yml
seirl: Yeah it's not really "to remove", it's more like "file upstream/distribution bugs because these aren't included and then remove"

PR #85, opened 2020-02-26
Add Rust generator

remi-dupre:
[Here is an example of how the API looks for prologin 2019](http://castor.dupre.io/drop/prologin2019/champion/index.html)

Issues
---

#### :warning: user functions returning heap-allocated data will leak memory

The only way to keep track of returned pointers would be to have some kind of global variable that would be allow to free data at next call to the API. I guess it would be doable but very ugly, and we'd have the same issue for all languages using C interface.

I'll have a look at C interface, I guess it could be simpler (and actually much more elegant) to be able to generate it with a move semantics (ie. it has to free every data it receives and forgets every data it sends). We should make sure to keep old behavior for C champion as I think it is more convenient to work with for a human.

#### CI tests may not pass

I think some dependancies is missing (I chose to link the same libraries as old generator without thinking about it). But it would be hard to investigate as Github CI seems to be down :'(

What changed?
---

#### use standard primitive types

I think `i32` is more convenient to work with than `c_int`, current generation might not convert to correct types on strange architectures but I don't think we have this kind of concern?

#### string are supported

Old API was providing/receiving a raw pointer `*const c_char` that candidates had to handle/free by themselves.

#### array parameters no longer leak memory

Old API was sending a raw pointers to C while loosing ownership.

#### more convenient parameters

 - no longer take `Copy` parameters by reference (no one wants a ref over integers :D)
 - take slices as parameter (that's a bit more generic than `Vec`, and more idiomatic)

#### more autotraits

When a struct does not contain a floating point number, it will implement `Eq`, `Hash` and `Ord`, this will help interactions with containers like `HashSet`, `BTreeMap`, ...

#### tuples are supported

I didn't even knew this feature existed :o


tools/generator/filters/rust.py
```diff
@@ -0,0 +1,77 @@
...
+        ret_annotation,
+    )
+
+
+@register_filter
+def rust_ffi_type(value: str) -> str:
+    generics = {
+        'bool': 'c_bool',
+        'double': 'c_double',
+        'float': 'c_float',
```
seirl: We just do doubles, not floats

tools/generator/filters/rust.py
```diff
@@ -0,0 +1,77 @@
+from functools import partial
+
+from . import register_filter
+from .common import generic_comment
+
+
+@register_filter
+def rust_call(func) -> str:
```
seirl: Doesn't seem rust specific, maybe put that in common? Also, can't you just do that in the templates directly?

tools/generator/filters/rust.py
```diff
@@ -0,0 +1,77 @@
...
+@register_filter
+def rust_prototype(func, ffi=False) -> str:
+    get_type = rust_ffi_type if ffi else rust_type
+
+    args = (
+        "{}: {}".format(arg_name, get_type(arg_type))
+        for arg_name, arg_type, _ in func['fct_arg']
+    )
+
+    if func["fct_ret_type"] == "void":
```
seirl: Use the common.returning() filter

tools/generator/wscript
```diff
@@ -59,6 +59,12 @@ def build(bld):
         'templates/player/cxx/Makefile.jinja2',
     ])
 
+    bld.install_files('${PREFIX}/lib/stechec2/generator/templates/player/rust', [
```
seirl: Rebase on master, this is no longer needed

tools/generator/filters/rust.py
```diff
@@ -0,0 +1,77 @@
...
+        prefix = value[:-len(" array")]
+        return "Vec<{}>".format(rust_type(prefix))
+    elif value in generics:
+        return generics[value]
+    else:
+        return value
+
+
+@register_filter
+def rust_comment(*args, doc: bool = False, **kwargs):
```
seirl: You can just do like for C here: import the cxx_comment function and register it.

tools/generator/macros/rust.jinja2
```diff
@@ -0,0 +1,40 @@
+{% macro decl_enum(enum) -%}
```
seirl: If these are only used once you don't need them as macros, just inline them in the templates directly.

tools/generator/templates/player/rust/api/ffi.rs.jinja2
```diff
@@ -0,0 +1,254 @@
...
+
+{% for struct in game.struct %}
+{% with ffi_type=struct.str_name|rust_ffi_type %}
+// Conversions for {{ struct.str_name }}
+
+{% with api_type=struct.str_name|rust_api_output_type(api_mod_path='api::') %}
+impl ToAPI<{{ api_type }}> for {{ ffi_type }} {
+    unsafe fn to_api(self) -> {{ api_type }} {
+        {% if struct.str_tuple %}
+        ({{ struct.str_field|map('first')|map('add_prefix', 'self.')|map('add_suffix', '.to_api()')|join(', ') }})
```
seirl: Use a for loop and `joiner()` instead of defining two filters just for that.

tools/generator/filters/common.py
```diff
@@ -31,6 +54,26 @@ def is_struct(ctx, type) -> bool:
     return is_struct_in(type, ctx['game'])
 
 
+@register_function
+@contextfunction
+def is_tuple(ctx, type) -> bool:
```
seirl: This should be a test, like "is_returning"

tools/generator/filters/common.py
```diff
@@ -31,6 +54,26 @@ def is_struct(ctx, type) -> bool:
...
+    struct = as_struct(ctx, type)
+    return struct is not None and struct['str_tuple']
+
+
+@register_filter
+@contextfilter
+def as_struct(ctx, type) -> Optional[dict]:
```
seirl: This is the kind of things that should probably be a get_struct function in `game`

remi-dupre: I can't find anything like `@contexttest` in jinja2 documentation, I've also tested that both `@contextfilter` and `@contextfunction` don't work together with `@register_test`.

tools/generator/filters/common.py
remi-dupre: I can't find anything like `@contexttest` in jinja2 documentation, I've also tested that both `@contextfilter` and `@contextfunction` don't work together with `@register_test`.

tools/generator/filters/common.py
```diff
@@ -16,21 +17,53 @@ def is_returning(func) -> bool:
...
+
+
+@register_test
+def is_tuple(struct) -> bool:
+    return struct['str_tuple']
+
+
+@register_filter
+@contextfilter
+def is_containing_double(ctx, type):
```
seirl: This is very specific, what about putting it in rust.py with a name like `rust_struct_has_equality`? (Focusing the interface name on what it *means* instead of the technical details).

tools/generator/filters/common.py
```diff
@@ -16,21 +17,53 @@ def is_returning(func) -> bool:
     return func['fct_ret_type'] != 'void'
 
 
+@register_function
+@contextfunction
+def is_enum(ctx, type) -> bool:
```
seirl: The idea was to move get_enum() and get_struct() in game.py to remove them from these files. They shouldn't be needed anymore.

tools/generator/filters/rust.py
```diff
@@ -0,0 +1,145 @@
...
+        ', '.join(args),
+        ret_annotation,
+    )
+
+
+@register_filter
+@contextfilter
+def rust_ffi_type(ctx, value: str) -> str:
+    """Type sent to the FFI"""
+    generics = {
```
seirl: Is "generics" a usual name for these? Isn't "base_types" better?

tools/generator/filters/common.py
```diff
@@ -16,21 +17,53 @@ def is_returning(func) -> bool:
...
+    return is_enum_in(type, ctx['game'])
+
+
+def is_enum_in(type, game) -> bool:
```
seirl: Remove

tools/generator/filters/common.py
```diff
@@ -16,21 +17,53 @@ def is_returning(func) -> bool:
...
-    return any(type == s['str_name'] for s in game['struct'])
-
-
 @register_function
 @contextfunction
 def is_struct(ctx, type) -> bool:
     return is_struct_in(type, ctx['game'])
 
 
+def is_struct_in(type, game) -> bool:
```
seirl: Remove

tools/generator/filters/common.py
```diff
@@ -16,21 +17,53 @@ def is_returning(func) -> bool:
...
     return type.endswith(' array')
 
 
-def is_struct_in(type, game) -> bool:
-    return any(type == s['str_name'] for s in game['struct'])
-
-
 @register_function
 @contextfunction
 def is_struct(ctx, type) -> bool:
```
seirl: Remove

tools/generator/filters/common.py
remi-dupre: Bim Bam Boum

https://github.com/prologin/stechec2/pull/85/commits/7311d0ab7943714dfbea6101ca1b7a4d8fe89804

tools/generator/game.py
```diff
@@ -165,6 +165,18 @@ def check_field_unicity(self):
                         .format(origin_type, origin, name, def_type, def_name))
                 used_field_names[name] = (origin_type, origin)
 
+    def get_struct(self, struct_name):
+        '''Get a struct by name, None if it does not exist'''
+        return next(
+            (
```
seirl, line 182: I think this pair of parens is useless

tools/generator/templates/includes/Makefile-language.jinja2
```diff
@@ -22,7 +22,7 @@ to_clean = $(objs)
 
 {% block build_library %}
 champion.so: $(objs)
-	$(CXX) -o $@ $^ -shared $(LDFLAGS)
+	$(CXX) -o $@ -Wl,--whole-archive $^ -Wl,--no-whole-archive -shared $(LDFLAGS)
```
seirl: Could you add a comment to explain this? Is that really something we want for all the languages as opposed to only for Rust?

tools/generator/player.py
```diff
@@ -45,3 +45,7 @@ def gen_lang(lang: str, *, files, symlinks) -> None:
     gen_lang('caml',
              files=['champion.ml', 'champion.mli', 'Makefile'],
              symlinks=['api.ml', 'interface.cc', 'Makefile-caml'])
+    gen_lang('rust',
+             files=['Makefile', 'Makefile-rust', 'Cargo.toml',
+                    'champion.rs', 'api/mod.rs', 'api/ffi.rs'],
```
seirl: A lot of these should be symlinks

tools/generator/templates/player/rust/Makefile-rust.jinja2
```diff
@@ -0,0 +1,23 @@
...
+RUSTC_FLAGS = --edition 2018 -g -C relocation-model=pic
+CPPFLAGS += -MMD -MP
+LDFLAGS = -lm -lrt -ldl -lpthread -lstdc++
+{% endblock %}
+
+{% block extra_rules %}
+objs += $(CHAMPION_TARGET).o
+
+$(CHAMPION_TARGET).o:
+	rustc --crate-type=staticlib $(RUSTC_FLAGS) --emit link,dep-info=.$(CHAMPION_TARGET).o.d $(CHAMPION_TARGET).rs -o $@
```
seirl: Shouldn't this output a .a file?

tools/generator/templates/player/rust/Makefile-rust.jinja2
```diff
@@ -0,0 +1,23 @@
...
+{% block lang %}rust{% endblock %}
+
+{% block extra_flags %}
+RUSTC_FLAGS = --edition 2018 -g -C relocation-model=pic
+CPPFLAGS += -MMD -MP
+LDFLAGS = -lm -lrt -ldl -lpthread -lstdc++
+{% endblock %}
+
+{% block extra_rules %}
+objs += $(CHAMPION_TARGET).o
```
seirl: Please hardcode "champion" for consistency, or change it in all the other languages

tools/generator/templates/player/rust/Makefile.jinja2
```diff
@@ -0,0 +1,15 @@
+{# SPDX-License-Identifier: GPL-2.0-or-later
+ # Copyright (c) 2020 Association Prologin <association@prologin.org>
+ #}
+# Add all your champion files in here, separated by spaces
+# Include all the sources (must end with *.c), headers (must end with *.h) and
+# README files.
+CHAMPION_TARGET = champion
```
seirl: The lib name shouldn't be something that can be changed here

tools/generator/templates/player/rust/api/ffi.rs.jinja2
```diff
@@ -0,0 +1,273 @@
...
+    {% for field_name, field_type, _ in struct.str_field %}
+    {{ field_name }}: {{ field_type|rust_ffi_type }},
+    {% endfor %}
+}
+
+{% endfor -%}
+
+// Conversion traits
+
+pub trait ToAPI<T> {
```
seirl: Do you think it could be a good idea to reuse the nomenclature used for the dynamic language interfaces here? Like CToRust and RustToC

seirl (changes requested):

tools/generator/templates/player/rust/api/ffi.rs.jinja2
remi-dupre: If I can help consistency I don't have bigger cons than "it takes a few more letters to write".

However I think that it is still more idiomatic to write `var.to_rust()` than `var.rust_to_c()`, maybe inner methods should keep this pattern.

tools/generator/filters/common.py
remi-dupre: More than equality it also affects inequality and hash.

But if we replace this by a rust-specific function I should probably just write a function which returns the list traits that should be auto-implemented given a type.

seirl: Sounds good

tools/generator/templates/player/rust/api/ffi.rs.jinja2
seirl: I understand but right now it's pretty hard to keep track of what is of what type, and maybe the explicit rust_to_c() could help

tools/generator/filters/rust.py
remi-dupre: Oh, I meant "primitives", sometime words get swapped in my head :(

remi-dupre: But that's actually not the correct word either since `str` and `[T]` are both primitives, let's go for `basic_types`

tools/generator/filters/rust.py
```diff
@@ -0,0 +1,174 @@
...
+    if is_array(value):
+        inherited = rust_auto_traits(ctx, get_array_inner(value))
+
+        if 'Copy' in inherited:
+            inherited.remove('Copy')
+
+        return inherited
+
+    if value == 'double':
+        return {'Copy', 'Clone', 'Debug', 'PartialEq', 'PartialOrd'}
```
seirl, line 177: Returning a set here might get you randomly ordered traits, which isn't great for gendiff etc. Why don't you return a list instead?

seirl (changes requested):

tools/generator/filters/rust.py
remi-dupre: That's because I am computing intersections over this result, that's why I am using the sort filter in the template. Is this still a big issue?

remi-dupre: That's because I am computing intersections over this result, that's why I am using the sort filter in the template. Is this still a big issue?

seirl: Ah, I hadn't seen the |sort filter, looks good then

tools/generator/game.py
remi-dupre: You can only omit them for single-argument function calls:

```python3
>>> next(i for i in range(10, 20), None)
  File "<stdin>", line 1
SyntaxError: Generator expression must be parenthesized if not sole argument
```

seirl (approved):

PR #90, opened 2020-02-28
generator: Java generator

Shaac:
Still missing in the PR:

- [x] Javadoc
- [x] Initialization of struct inside struct (add a `is_struct` test)
- [x] make (dist)clean of *.class

### DIFF
--- a/tools/generator/filters/java.py
+++ b/tools/generator/filters/java.py
@@ -1,0 +1,141 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+# Copyright (c) 2020 Association Prologin <association@prologin.org>
+# Copyright (c) 2020 Sacha Delanoue
+# Copyright (c) 2020 Antoine Pietri
+
+from functools import partial
+
+from . import register_filter
+from .common import camel_case, generic_args, generic_comment
+from .common import generic_prototype, get_array_inner, is_array
+from .cxx import cxx_type
+
+
+@register_filter
+def java_type(value: str) -> str:
+    """Get the type name in the Java language"""
+    base_types = {
+        'void': 'void',
+        'int': 'int',
+        'double': 'double',
+        'bool': 'boolean',
+        'string': 'String',
+    }
+    if value in base_types:
+        return base_types[value]
+    if is_array(value):
+        return "{}[]".format(java_type(get_array_inner(value)))
+    return camel_case(value)
+
+
+@register_filter
+def java_args(value, type_mapper=lambda x: x) -> str:
+    return generic_args(value, type_mapper=java_type)
+
+
+java_prototype = register_filter(
+    partial(generic_prototype, type_mapper=java_type, arg_mapper=java_args),
+    name='java_prototype',
+)
+
+
+@register_filter
+def java_comment(value: str, doc: bool = False, indent: int = 0) -> str:
+    if not doc:
+        return generic_comment(value, "// ", indent)
+    return "/**\n" + indent * " " + generic_comment(
+        value, " * ", indent) + "\n" + indent * " " + " */"
+
+
+def _jtype(typename: str) -> str:
+    """Return the type to be used as for java in java_to_cxx and cxx_to_java"""
+    generics = {
+        'bool': 'jboolean',
+        'double': 'jdouble',
+        'int': 'jint',
+        'string': 'jstring',
+    }
+    if typename in generics.keys():
+        return generics[typename]
+    return camel_case(typename)
+
+
+@register_filter
+def java_to_cxx(value: str) -> str:
+    if is_array(value):
+        base_type = get_array_inner(value)
+        return 'java_to_cxx_array<{}, {}>'.format(
+            _jtype(base_type),
+            cxx_type(base_type)
+        )
+    elif value in ('int', 'double', 'bool', 'string'):
+        return 'java_to_cxx<{}, {}>'.format(_jtype(value), cxx_type(value))
+    else:
+        return 'java_to_cxx<jobject, {}>'.format(cxx_type(value))
+
+
+@register_filter
+def cxx_to_java(value: str) -> str:
+    if is_array(value):
+        base_type = get_array_inner(value)
+        return 'cxx_to_java_array<{}, {}>'.format(
+            cxx_type(base_type),
+            _jtype(base_type)
+        )
+    elif value in ('int', 'double', 'bool', 'string'):
+        return 'cxx_to_java<{}, {}>'.format(cxx_type(value), _jtype(value))
+    else:
+        return 'cxx_to_java<{}, jobject>'.format(cxx_type(value))
+
+
+@register_filter
+def jni_type(value: str) -> str:
+    """Return the real JNI type"""
+    if value in ('boolean', 'int', 'double'):
+        return 'j' + value
+    elif value == 'bool':
+        return "jboolean"
+    elif is_array(value):
+        return "jarray"
+    elif value == "void" or not value:
+        return "void"
+    return "jobject"
+
+
+@register_filter
+def jni_type_signature(value: str) -> str:
+    """Get Java VM's representation of type signature for a type"""
+    if is_array(value):
+        return "[" + jni_type_signature(get_array_inner(value))
+    dic = {
+        'bool': 'Z',
+        'double': 'D',
+        'int': 'I',
+        'void': 'V',
+        'string': 'Ljava/lang/String;',
+    }
+    if value in dic.keys():
+        return dic[value]
+    return 'L' + camel_case(value) + ';'
+
+
+@register_filter
+def jni_method_signature(func) -> str:
+    """Get Java VM's representation of type signature for a method"""
+    return "({}){}".format(
+        "".join(
+            jni_type_signature(arg_type)
+            for _, arg_type, _ in func['fct_arg']),
+        jni_type_signature(func['fct_ret_type']))
+
+
+@register_filter
+def java_type_method_call(value: str) -> str:
+    """return type used in the Call<TYPE>Method functions (and alike)"""
+    if value in ('int', 'double'):
+        return value.capitalize()
+    elif value == 'bool':
+        return "Boolean"
+    elif value == 'void' or not value:
+        return "Void"
+    return "Object"
--- a/tools/generator/templates/player/java/interface.cc.jinja2
+++ b/tools/generator/templates/player/java/interface.cc.jinja2
@@ -1,0 +1,347 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+// Copyright (c) 2005-2020 Association Prologin <association@prologin.org>
+{% import 'macros/cxx.jinja2' as cxx %}
+
+// This file contains the code to call the API functions from the Java language.
+// {{ stechec2_generated }}
+
+#include <cstdlib>
+#include <string>
+#include <vector>
+
+#include <jni.h>
+
+struct Champion
+{
+    static jclass Class();
+};
+{% for enum in game.enum %}
+
+{{ cxx.decl_enum(enum) }}
+
+struct {{ enum.enum_name|camel_case }}
+{
+    static jclass Class();
+};
+{% endfor %}
+{% for struct in game.struct %}
+
+{{ cxx.decl_struct(struct) }}
+
+struct {{ struct.str_name|camel_case }}
+{
+    static jclass Class();
+};
+{% endfor %}
+
+struct StechecJavaRunTime
+{
+    StechecJavaRunTime() = default;
+    ~StechecJavaRunTime();
+    void init();
+    bool function_enter();
+    void function_exit(bool attached);
+
+    JavaVM* jvm;
+    JNIEnv* env;
+    jobject prologin;
+
+private:
+    bool is_init_ = false;
+};
+static struct StechecJavaRunTime jrt;
+
+{% for func in game.function %}
+extern "C" {{ func|cxx_prototype(prefix='api_') }};
+{% endfor %}
+{% for enum in game.enum %}
+extern "C" void api_afficher_{{ enum.enum_name }}({{ enum.enum_name }} v);
+{% endfor %}
+{% for struct in game.struct %}
+extern "C" void api_afficher_{{ struct.str_name }}({{ struct.str_name }} v);
+{% endfor %}
+
+// In case of errors...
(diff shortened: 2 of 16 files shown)


tools/generator/filters/java.py
```diff
@@ -0,0 +1,131 @@
...
+from functools import partial
+
+from . import register_filter
+from .common import camel_case, generic_args, generic_prototype, is_array
+from .cxx import cxx_comment, cxx_type
+
+
+@register_filter
+def java_type(value: str) -> str:
+    if value in ('byte', 'char', 'short', 'int', 'long', 'float', 'double', 'void'):
```
seirl: I think we should only handle the types stechec already handles to avoid confusion

tools/generator/filters/java.py
```diff
@@ -0,0 +1,131 @@
...
+@register_filter
+def get_fn_signature(func) -> str:
+    return "(" + "".join(get_signature(arg_type) for _, arg_type, _ in func['fct_arg']) + ")" + get_signature(func['fct_ret_type'])
+
+
+@register_filter
+def get_java_type(value: str) -> str:
+    """return the type to be used in the Call<TYPE>Method functions (and alike)
+    """
+    if value in ('byte', 'char', 'short', 'int', 'long', 'float', 'double'):
```
seirl: ditto

tools/generator/filters/java.py
```diff
@@ -0,0 +1,131 @@
...
+        return "Boolean"
+    elif value == 'void' or not value:
+        return "Void"
+    return "Object"
+
+
+@register_filter
+def get_jni_type(value: str) -> str:
+    """return the real JNI type"""
+    if value in ('boolean', 'byte', 'char', 'short', 'int', 'long', 'float', 'double', 'string'):
```
seirl: ditto

tools/generator/filters/java.py
```diff
@@ -0,0 +1,131 @@
...
+        'double': 'jdouble',
+        'int': 'jint',
+        'string': 'jstring',
+    }
+    if typename in generics.keys():
+        return generics[typename]
+    return camel_case(typename)
+
+
+def _cxx_type(value: str) -> str:
```
seirl: Could you explain the need for this?

tools/generator/filters/java.py
```diff
@@ -0,0 +1,131 @@
...
+        'string': 'Ljava/lang/String;',
+    }
+    if value in dic.keys():
+        return dic[value]
+    return 'L' + camel_case(value) + ';'
+
+
+@register_filter
+def get_fn_signature(func) -> str:
+    return "(" + "".join(get_signature(arg_type) for _, arg_type, _ in func['fct_arg']) + ")" + get_signature(func['fct_ret_type'])
```
seirl: PEP8

Please add docstrings to explain the difference between a signature and a prototype?

tools/generator/filters/java.py
Shaac: done

tools/generator/filters/java.py
Shaac: done

tools/generator/filters/java.py
Shaac: done

tools/generator/filters/java.py
Shaac: done

tools/generator/filters/java.py
Shaac: I renamed the function and added a comment.

Shaac:
I’d still like to convert the comments in Api.java and Champion.java to real JavaDoc, but apart from that I’m pretty satisfied.

tools/generator/filters/common.py
```diff
@@ -57,3 +62,9 @@ def generic_comment(value: str, start: str, indent: int = 0) -> str:
         break_long_words=False,
         replace_whitespace=False
     ))
+
+
+@register_filter
+def get_cst_type(constant) -> type:
```
seirl: I think this should go in `game.py`, we shouldn't have to let the templates do this.

tools/generator/filters/java.py
```diff
@@ -0,0 +1,141 @@
...
+
+from . import register_filter
+from .common import camel_case, generic_args, generic_prototype, is_array
+from .cxx import cxx_comment, cxx_type
+
+
+@register_filter
+def java_type(value: str) -> str:
+    """Get the type name in the Java language"""
+    if value in ('int', 'double', 'void'):
```
seirl: It looks like it would be prettier with just:

```
base_types = {
    'void': 'void',
    'int': 'int',
    'double': 'double',
    'string': 'String',
}
if value in base_types:
     return base_types[value]
else ...

File: tools/generator/filters/java.py (hunk 4 above)
Comment (seirl) [reply tools/generator/filters/java.py]: Sorry, this is still a bit unclear for me just reading this without looking at the context: why do we need two kinds of C++ types (bool and boolean)? Why isn't this just abstracted away in the `java_to_cxx` functions?

File: tools/generator/templates/player/java/interface.cc.jinja2 [hunk 8]
```diff
@@ -0,0 +1,338 @@
...
+{% endfor %}
+{% for struct in game.struct %}
+extern "C" void api_afficher_{{ struct.str_name }}({{ struct.str_name }} v);
+{% endfor %}
+
+// In case of errors...
+template <typename Lang, typename CxxType>
+CxxType java_to_cxx(Lang in)
+{
+    return in.error_should_not_happens;
```
seirl: should not happen

tools/generator/templates/player/java/interface.cc.jinja2
```diff
@@ -0,0 +1,338 @@
...
+
+template<typename Lang, typename CxxType>
+std::vector<CxxType> java_to_cxx_array(jobject in);
+
+template<typename CxxType, typename Lang>
+jarray cxx_to_java_array(std::vector<CxxType> in);
+
+// Basic type wrappers
+{% for type in ['boolean', 'int', 'double'] %}
+{% set jnitype = 'j' + type %}
```
seirl: Isn't this obsoleted by |jni_type?

tools/generator/test/games/test.yml
```diff
@@ -39,6 +39,13 @@ struct:
       - [field_int, int, "Integer field"]
       - [field_int_arr, int array, "Integer array field"]
       - [field_str_arr, simple_struct array, "Structure array field"]
+  -
```
seirl, line 42: Maybe it's better to have this in an other PR, since it might break the CI for other languages for something not linked to Java?

tools/generator/filters/java.py
Shaac: Two reasons why it is done at this level:
The first one is that it is how the ruby generator does it.
The second, is that, because of lines like `jrt.env->SetBooleanArrayRegion(out, (jsize)0, (jsize)in.size(), (const jboolean*)in.data());`, if `in` is a `std::vector<bool>`, then `.data()` is not a `bool*` and then the entire behavior changes.

So is it possible to change the behavior so that we see only `bool` and not `boolean` in the C++? Yes. Is it fast and simple? No.

tools/generator/filters/common.py
Shaac: Moved the function as a member of Game.

tools/generator/filters/java.py
Shaac: Done.

tools/generator/templates/player/java/interface.cc.jinja2
Shaac: done

tools/generator/templates/player/java/interface.cc.jinja2
Shaac: Well it’s more or less the same. Is `type|jni_type` better than `'j' + type`?

tools/generator/test/games/test.yml
Shaac: I’d agree if it did break the CI. But since it does not (and frankly, the yaml already generate a struct of list of struct, here it is just a struct of struct, so I don’t see how it would break stuff), I don’t really see the point in making an other PR.

Here it is just to see the codepath in the java generator especially made for struct of struct.

tools/generator/templates/player/java/interface.cc.jinja2
seirl: I think the semantics is clearer and avoids the need for the intermediate {% set

tools/generator/templates/player/java/interface.cc.jinja2
```diff
@@ -0,0 +1,338 @@
...
+    jrt.env->Set{{ type|capitalize }}ArrayRegion(out, (jsize)0, (jsize)in.size(), (const {{ jnitype }}*)in.data());
+    return (jarray)out;
+}
+
+template <>
+std::vector<{{ type }}> java_to_cxx_array<{{ jnitype }}, {{ type }}>(jobject in)
+{
+    {{ arraytype }} array = ({{ arraytype }})in;
+    jsize size = jrt.env->GetArrayLength(array);
+    {{jnitype}}* datas = jrt.env->Get{{ type|capitalize }}ArrayElements(array, NULL);
```
seirl: Spacing around {{

tools/generator/templates/player/java/interface.cc.jinja2
```diff
@@ -0,0 +1,338 @@
...
+
+template<>
+jobject cxx_to_java<{{ enum.enum_name }}, jobject>({{ enum.enum_name }} in)
+{
+    jmethodID method = jrt.env->GetStaticMethodID({{ enum.enum_name|camel_case }}::Class(), "values", "()[L{{ enum.enum_name|camel_case }};");
+    jobjectArray values = (jobjectArray)jrt.env->CallStaticObjectMethod({{ enum.enum_name|camel_case }}::Class(), method);
+    return jrt.env->GetObjectArrayElement(values, (jsize)in);
+}
+{% endfor %}
+{% for struct in game.struct %}
```
seirl, line 185: Blank newline here maybe?

tools/generator/templates/player/java/interface.cc.jinja2
```diff
@@ -0,0 +1,338 @@
...
+    (api_{{ func.fct_name }}(
+    {%- set comma = joiner(', ') -%}
+    {%- for arg_name, arg_type, _ in func.fct_arg -%}
+    {{ comma() }}{{ arg_type|java_to_cxx }}({{ arg_name }})
+    {%- endfor %}));
+}
+{% endfor %}
+{% for enum in game.enum %}
+
+void afficher_{{ enum.enum_name }}(JNIEnv* _env, jobject _obj, jobject v)
```
seirl: Remove the unused argument names

tools/generator/templates/player/java/interface.cc.jinja2
```diff
@@ -0,0 +1,338 @@
...
+{
+    {%+ if func is returning %}return {{ func.fct_ret_type|cxx_to_java }}{% endif -%}
+    (api_{{ func.fct_name }}(
+    {%- set comma = joiner(', ') -%}
+    {%- for arg_name, arg_type, _ in func.fct_arg -%}
+    {{ comma() }}{{ arg_type|java_to_cxx }}({{ arg_name }})
+    {%- endfor %}));
+}
+{% endfor %}
+{% for enum in game.enum %}
```
seirl, line 222: I think we said we were going to generate the afficher_* functions directly in game.py

tools/generator/templates/player/java/interface.cc.jinja2
Shaac: done

tools/generator/templates/player/java/interface.cc.jinja2
Shaac: done

tools/generator/templates/player/java/interface.cc.jinja2
Shaac: I agree, but while it is not done at least the generator is complete.

tools/generator/templates/player/java/interface.cc.jinja2
Shaac: done

tools/generator/templates/player/java/interface.cc.jinja2
Shaac: I remove the `jnitype` variable.

tools/generator/filters/java.py
Shaac: Ok I removed the whole `boolean` stuff. Now it uses `bool` for C++, and it works.

tools/generator/templates/player/java/interface.cc.jinja2
```diff
@@ -0,0 +1,345 @@
...
+{{ type }} java_to_cxx<{{ type|jni_type }}, {{ type }}>({{ type|jni_type }} in)
+{
+    return ({{ type }})in;
+}
+
+template <>
+jarray cxx_to_java_array<{{ type }}, {{ type|jni_type }}>(std::vector<{{ type }}> in)
+{
+    {{ arraytype }} out = jrt.env->New{{ type|java_type_method_call }}Array((jsize)in.size());
+    {% if type != 'bool' %}
```
seirl, line 102: Maybe add a comment here to explain?

tools/generator/templates/player/java/interface.cc.jinja2
```diff
@@ -0,0 +1,345 @@
...
+template<typename CxxType, typename Lang>
+jarray cxx_to_java_array(std::vector<CxxType> in);
+
+// Basic type wrappers
+{% for type in ['bool', 'int', 'double'] %}
+{% set arraytype = type|jni_type + 'Array' %}
+template <>
+{{ type|jni_type }} cxx_to_java<{{ type }}, {{ type|jni_type }}>({{ type }} in)
+{
+    return ({{ type|jni_type }})in;
```
seirl, line 89: I think we put spaces after casts in our format options

Shaac: I ran clang-format to see and we don’t put spaces.

tools/generator/templates/player/java/interface.cc.jinja2
Shaac: done

seirl (approved):

PR #91, opened 2020-02-29
generator: refactor makefiles with inheritance

seirl:


tools/generator/templates/includes/Makefile.language.jinja2
```diff
@@ -0,0 +1,48 @@
...
+
+to_clean :=
+
+cxx_sources = $(filter %.cc,$(CHAMPION_FILES))
+objs += $(cxx_sources:.cc=.o)
+
+{% block extra_rules %}
+{% endblock %}
+
+deps = $(objs:.o=.d)
```
Shaac: I don’t think every .o obtained by any language has a corresponding .d file.

I don’t like that the dependency is done on the main makefile. the -MMD flags is specific to C/C++, and should be present for other languages. Each language will have to manage their own dependencies anyway.

tools/generator/templates/includes/Makefile.language.jinja2
```diff
@@ -0,0 +1,48 @@
+# SPDX-License-Identifier: GPL-2.0-or-later
+# Copyright (c) 2020 Association Prologin <association@prologin.org>
+
+# Add all your champion files in here, separated by spaces
+# Include all the sources, headers and README files.
```
Shaac, line 5: I’m affraid if we don’t specify the extention here, the candidate may use extensions like .cpp because they prefere it, and the makefile wo’nt handle those.

PR #92, opened 2020-03-01
generator: split makefiles in an editable and a non-editable file

seirl:


Shaac (commented):
It seems unnecessarily complicated to have 2 makefiles for each langage. Can’t this work with just one makefile?

Shaac (approved):

seirl:
As discussed on IRC, this is necessary to symlink the core content of the makefiles.

PR #94, opened 2020-03-01
add missing include in rules/src/entry.cc

remi-dupre:
This include is required at [line 22](https://github.com/prologin/stechec2/compare/missing-rule-include?expand=1#diff-e1d7e9f064676a754e1c653c826b4c44L22) to get definition of `rules::Config`.


fnareoh (approved):
He committed but I'm the one who found it :p

Issue #45, opened 2019-06-01
Recode all the generators in Jinja


seirl:
## Misc generators

- [x] Règles (@Shaac) : #66 
- [x] Sphinx (@seirl) : c6122bde16064975062d86133527c279375b3895
- [x] Texdoc (@seirl) : 8b7542a2147839e06573702aee5a2294643b066c

## Champions

- [x] C++ (@Shaac) : #75 
- [x] C (@seirl) : #76 
- [x] Python (@seirl) : #83 
- [x] PHP (@seirl) : #84 
- [x] Java (@Shaac) : #90 
- [x] OCaml (@Dettorer) : #97
- [x] C# (@alarsyo) : #110 
- [x] Haskell (@zuh0) : #98 
- [x] Rust (@remi-dupre) : #85 

## Other

- [x] Test loading (@seirl) : #80 
- [x] Test champions (@seirl) : c09a648
- [x] Server makefiles (@seirl) : #92 
- [x] Symlinks (@seirl) : #93

## Minor issues
- [ ] Inline api.h in the C API so that the `interface.cc` is self-sufficient for languages that depend on it
- [x] Load the correct `cst_type` directly when loading the constants in `game.py` instead of having to call a helper (#114)
- [x] Load the `afficher_*` functions (with a special attribute like `fct_display: true` so that we can filter them out if necessary) (#113)
- [x] Filter out reserved language keywords while loading the YAML (#109)

alarsyo:
Could you add me on the C# and Go champions generators ? :)

PR #97, opened 2020-03-01
generator: add OCaml generator

Dettorer:
This only try to replicate the old ruby generator behavior. Some choices
need to be reconsidered in the future.

Advances #45

### DIFF
--- a/tools/generator/filters/__init__.py
+++ b/tools/generator/filters/__init__.py
@@ -30,10 +30,10 @@ def load_library_in(env):
     from . import (  # noqa: E402, F401
         common,
         c,
+        caml,
         cxx,
         php,
-        python,
-        caml
+        python
     )
     for filter_name, filter in FILTER_LIBRARY.items():
         env.filters[filter_name] = filter
--- a/tools/generator/filters/__init__.py
+++ b/tools/generator/filters/__init__.py
@@ -32,7 +32,8 @@ def load_library_in(env):
         c,
         cxx,
         php,
-        python
+        python,
+        caml
     )
     for filter_name, filter in FILTER_LIBRARY.items():
         env.filters[filter_name] = filter
--- a/tools/generator/filters/caml.py
+++ b/tools/generator/filters/caml.py
@@ -13,14 +13,6 @@ def caml_type(value: str) -> str:
     return value
 
 
-@register_filter
-def caml_callback(func) -> str:
-    '''symbolic name of a function to be used with Callback.register in caml
-    code, and with caml_named_value in C code
-    '''
-    return 'ml_' + func['fct_name']
-
-
 # caml functions need to take at least one parameter, be it a unit value if
 # there is nothing useful to give
 
--- a/tools/generator/filters/caml.py
+++ b/tools/generator/filters/caml.py
@@ -25,7 +25,7 @@ def caml_callback(func) -> str:
 # there is nothing useful to give
 
 @register_filter
-def cxx_nonempty_or_unit(arg_list):
+def caml_cxx_args(arg_list):
     if not arg_list:
         return [('unit', 'void', 'empty caml parameter')]
     return arg_list
--- a/tools/generator/filters/caml.py
+++ b/tools/generator/filters/caml.py
@@ -1,0 +1,86 @@
+import textwrap
+
+from . import register_filter
+from .common import is_array
+
+
+@register_filter
+def caml_type(value: str) -> str:
+    if value == 'void':
+        return 'unit'
+    if value == 'double':
+        return 'float'
+    return value
+
+
+@register_filter
+def caml_callback(func) -> str:
+    '''symbolic name of a function to be used with Callback.register in caml
+    code, and with caml_named_value in C code
+    '''
+    return 'ml_' + func['fct_name']
+
+
+# caml functions need to take at least one parameter, be it a unit value if
+# there is nothing useful to give
+
+@register_filter
+def cxx_nonempty_or_unit(arg_list):
+    if not arg_list:
+        return [('unit', 'void', 'empty caml parameter')]
+    return arg_list
+
+
+@register_filter
+def caml_prototype(func) -> str:
+    arg_list = ' '.join(arg_name for arg_name, _, _ in func['fct_arg'])
+    if arg_list == '':  # make unit the only parameter
+        arg_list = '()'
+
+    return 'let {} {} ='.format(func['fct_name'], arg_list)
+
+
+@register_filter
+def caml_signature(func, external: bool = False) -> str:
+    arg_list = ' -> '.join(caml_type(arg_type)
+                           for _, arg_type, _
+                           in func['fct_arg'])
+    if arg_list == '':  # make unit the only parameter
+        arg_list = 'unit'
+
+    return '{} {} : {} -> {}'.format('external' if external else 'val',
+                                     func['fct_name'],
+                                     arg_list,
+                                     caml_type(func['fct_ret_type']))
+
+
+@register_filter
+def caml_comment(value: str, indent: int = 0, doc: bool = False) -> str:
+    # Make sure start and multiline_start are the same length
+    start = '(** ' if doc else '(* '
+    multiline_start = ' ** ' if doc else '** '
+    newline = "\n" + indent * " " + multiline_start
+
+    return start + newline.join(textwrap.wrap(
+        value + ' *)',
+        79 - indent - len(multiline_start),
+        expand_tabs=False,
+        break_long_words=False,
+        replace_whitespace=False
+    ))
+
+
+@register_filter
+def cxx_to_caml(caml_type: str) -> str:
+    if is_array(caml_type):
+        return 'cxx_to_caml_array'
+    else:
+        return 'cxx_to_caml<value, {}>'.format(caml_type)
+
+
+@register_filter
+def caml_to_cxx(value: str) -> str:
+    if is_array(value):
+        return 'caml_to_cxx_array<{}>'.format(value[:-len(' array')])
+    else:
+        return 'caml_to_cxx<value, {}>'.format(value)
--- a/tools/generator/player.py
+++ b/tools/generator/player.py
@@ -29,3 +29,6 @@ def make_player(game, out_dir: Path) -> None:
     gen_lang(game, out_dir, 'python',
              ['api.py', 'Champion.py', 'interface.cc', 'Makefile',
               'Makefile-python'])
+    gen_lang(game, out_dir, 'caml',
+             ['api.ml', 'champion.ml', 'champion.mli', 'interface.cc',
+              'Makefile', 'Makefile-caml'])
--- a/tools/generator/templates/player/caml/Makefile-caml.jinja2
+++ b/tools/generator/templates/player/caml/Makefile-caml.jinja2
@@ -3,7 +3,7 @@
 {% block lang %}caml{% endblock %}
 
 {% block extra_flags %}
-OCAMLC = ocamlc
+OCAMLC ?= ocamlc
 OCAML_PATH = $(shell $(OCAMLC) -where)
 OCAML_FLAGS = -g
 
--- a/tools/generator/templates/player/caml/Makefile-caml.jinja2
+++ b/tools/generator/templates/player/caml/Makefile-caml.jinja2
@@ -1,0 +1,37 @@
+{% extends "includes/Makefile-language.jinja2" %}
+
+{% block lang %}caml{% endblock %}
+
+{% block extra_flags %}
+OCAMLC = ocamlc
+OCAML_PATH = $(shell $(OCAMLC) -where)
+OCAML_FLAGS = -g
+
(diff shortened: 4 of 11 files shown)


tools/generator/filters/__init__.py
```diff
@@ -32,7 +32,8 @@ def load_library_in(env):
         c,
         cxx,
         php,
-        python
+        python,
+        caml
```
seirl: Alphabetic ordering

tools/generator/filters/caml.py
```diff
@@ -0,0 +1,87 @@
...
+    code, and with caml_named_value in C code
+    '''
+    return 'ml_' + func['fct_name']
+
+
+# caml functions need to take at least one parameter, be it a unit value if
+# there is nothing useful to give
+
+@register_filter
+def cxx_nonempty_or_unit(arg_list):
```
seirl: Why `cxx_`?

tools/generator/filters/caml.py
```diff
@@ -0,0 +1,87 @@
...
+    if not arg_list:
+        return [('unit', 'void', 'empty caml parameter')]
+    return arg_list
+
+
+@register_filter
+def caml_prototype(func) -> str:
+    arg_list = ' '.join(arg_name for arg_name, _, _ in func['fct_arg'])
+    if arg_list == '':  # make unit the only parameter
```
seirl, line 29: Can't you just pass your arglist through `nonempty_or_unit` first to avoid specialcasing?

tools/generator/player.py
```diff
@@ -26,3 +26,6 @@ def make_player(game, out_dir: Path) -> None:
              ['api.php', 'champion.php', 'interface.cc', 'Makefile'])
     gen_lang(game, out_dir, 'python',
              ['api.py', 'Champion.py', 'interface.cc', 'Makefile'])
+    gen_lang(game, out_dir, 'caml',
+             ['api.ml', 'champion.ml', 'champion.mli', 'interface.cc',
+              'Makefile'])
```
seirl: Rebase on master, this changed

tools/generator/templates/player/caml/interface.cc.jinja2
```diff
@@ -0,0 +1,236 @@
...
+extern "C" value ml_{{ func.fct_name }}({# argument list #}
+                                        {%- set comma = joiner(', ') -%}
+                                        {%- for arg_name, _, _ in arg_list -%}
+                                        {{ comma() }}value {{ arg_name }}
+                                        {%- endfor -%}
+                                       )
+{
+    CAMLparam0();
+    {# TODO: forbid instanciation if there's more than 5 parameters, OCaml
+     # has CAMLxparam0 to CAMLxparam5 only #}
```
seirl: TODO should be "Use CAMLparamN, then CAMLxparamN, by grouping the arguments 5 by 5"

tools/generator/templates/player/caml/interface.cc.jinja2
```diff
@@ -0,0 +1,236 @@
...
+template <>
+value cxx_to_caml<value, int>(int in)
+{
+    CAMLparam0();
+    CAMLreturn(Val_int(in));
+}
+template <>
+value cxx_to_caml<value, double>(double in)
+{
+    CAMLparam0();
```
seirl, line 50: Investigate whether these are necessary in the internal conversion functions, it seems weird.

Dettorer: That's the old generator's behavior, but it's something we should investigate yes. I'll add a TODO.

tools/generator/filters/__init__.py
Dettorer: It's already not in alphabetic ordering, should I sort the whole list or just insert caml after c?

Dettorer: Oh no I see, common should still be first?

seirl: Right.

tools/generator/filters/caml.py
Dettorer: It's used in parameters list in C++ functions, see https://github.com/prologin/stechec2/pull/97/files#diff-0ab55a5f02d7d1a0896ff290cce05feeR177.

tools/generator/filters/caml.py
Dettorer: I would need to make an other filter to handle empty parameter list in ocaml functions.

tools/generator/filters/caml.py
seirl: Okay, but this still gets added to the global scope available to all languages, so you should at least mention ocaml in the name

tools/generator/filters/caml.py
```diff
@@ -0,0 +1,86 @@
...
+def caml_type(value: str) -> str:
+    if value == 'void':
+        return 'unit'
+    if value == 'double':
+        return 'float'
+    return value
+
+
+@register_filter
+def caml_callback(func) -> str:
```
seirl: Just inline this honestly, it's really not worth a filter

tools/generator/templates/player/caml/Makefile.jinja2
```diff
@@ -0,0 +1,11 @@
+{# SPDX-License-Identifier: GPL-2.0-or-later
+ # Copyright (c) 2020 Association Prologin <association@prologin.org> #}
+# Add all your champion files in here, separated by spaces
+# Include all the sources (must end with *.py) and README files.
```
seirl: .ml here

tools/generator/templates/player/caml/api.ml.jinja2
```diff
@@ -0,0 +1,39 @@
+{# Constants #}
+{% for constant in game.constant %}
+{{ constant.cst_comment|caml_comment(doc=True) }}
+let {{ constant.cst_name|lower }} = {{ constant.cst_val }}
+
```
seirl, line 5: Generally we put the newline after the for, so that it avoids leaving empty lines afterward (especially useful at the end of the file)

tools/generator/templates/player/caml/api.ml.jinja2
```diff
@@ -0,0 +1,39 @@
...
+  {% for field_name, field_type, field_comment in struct.str_field %}
+  {{ field_name }} : {{ field_type|caml_type }}; (** <- {{ field_comment }} *)
+  {% endfor %}
+}
+{% endif %}
+
+{% endfor %}
+{% for func in game.function %}
+{{ func.fct_summary|caml_comment(doc=True) }}
+{{ func|caml_signature(external=True) }} = "ml_{{ func.fct_name }}"
```
seirl, line 37: Yeah, you even already inlined it here, good :-)

tools/generator/templates/player/caml/interface.cc.jinja2
```diff
@@ -0,0 +1,238 @@
...
+
+extern "C" {
+{% for func in game.function %}
+
+{{ func.fct_summary|cxx_comment(doc=True) }}
+{{ func|cxx_prototype(prefix='api_') }};
+{% endfor %}
+}
+
+template <typename CamlType, typename Cxx>
```
seirl: We use `CxxType` in the other generators

tools/generator/templates/player/caml/interface.cc.jinja2
```diff
@@ -0,0 +1,238 @@
...
+    CAMLparam0();
+    CAMLreturn(caml_copy_double(in));
+}
+template<>
+value cxx_to_caml<value, std::string>(std::string in)
+{
+    CAMLparam0();
+    size_t l = in.length();
+    char * out = (char *) malloc(l + 1);
+    for (size_t i = 0; i < l; i++) out[i] = in[i];
```
seirl: Could you try to fix the formatting so that it matches our general C++ style? here you'd need a newline

tools/generator/templates/player/caml/interface.cc.jinja2
```diff
@@ -0,0 +1,238 @@
...
+{
+    CAMLparam0();
+    CAMLreturn(caml_copy_double(in));
+}
+template<>
+value cxx_to_caml<value, std::string>(std::string in)
+{
+    CAMLparam0();
+    size_t l = in.length();
+    char * out = (char *) malloc(l + 1);
```
seirl: `char* out`

tools/generator/templates/player/caml/interface.cc.jinja2
```diff
@@ -0,0 +1,238 @@
...
+    for (size_t i = 0; i < l; i++) out[i] = in[i];
+    out[l] = 0;
+    CAMLreturn(caml_copy_string(out));
```
seirl: caml_copy_string is already doing a strdup. Why isn't this function just a `CAMLreturn(caml_copy_string(in.c_str()));` ?

tools/generator/templates/player/caml/interface.cc.jinja2
```diff
@@ -0,0 +1,238 @@
...
+}
```
seirl, line 62: Newline between functions

tools/generator/templates/player/caml/interface.cc.jinja2
seirl: (If it's different than that, you probably want to free the string before the return.)

tools/generator/templates/player/caml/interface.cc.jinja2
```diff
@@ -0,0 +1,238 @@
...
+
+{{ struct.str_summary|cxx_comment }}
+template<>
+value cxx_to_caml<value, {{ struct.str_name }}>({{ struct.str_name }} in)
+{
+    CAMLparam0();
+    CAMLlocal1(out);
+    out = caml_alloc({{ struct.str_field|length }}, 0);
+    {% for field_name, field_type, field_comment in struct.str_field %}
+    caml_initialize(&Field(out, {{ loop.index0 }}), {{ field_type|cxx_to_caml}}(in.{{ field_name }}));
```
seirl: Space before }}

tools/generator/templates/player/caml/interface.cc.jinja2
```diff
@@ -0,0 +1,238 @@
...
+** Inititialize caml
+*/
+static inline void _init_caml()
+{
+    static bool is_initialized = false;
+    if (!is_initialized)
+    {
+        is_initialized = true;
+        const char* argv[2] = {"./caml", NULL};
+        caml_startup(const_cast<char**>(argv));
```
seirl: This is not great. Consider using:

```
char program_name[] = "./caml";
char* argv[2] = {program_name, NULL};
caml_startup(argv);
```

Yes, it's annoying, but it's the proper way to declare what you want (allocate the string on the stack then deallocate it after the startup, which can change its arguments).

tools/generator/templates/player/caml/api.ml.jinja2
Dettorer: I know, but here it would add a empty line at the start of the for.

tools/generator/templates/player/caml/api.ml.jinja2
Dettorer: No it wasn't the same goal here, it's not a OCaml symbol that we need to grab from C++.

tools/generator/templates/player/caml/Makefile-caml.jinja2
```diff
@@ -0,0 +1,37 @@
+{% extends "includes/Makefile-language.jinja2" %}
+
+{% block lang %}caml{% endblock %}
+
+{% block extra_flags %}
+OCAMLC = ocamlc
```
seirl: `?=`

tools/generator/templates/player/caml/interface.cc.jinja2
Dettorer: That was in the old generator, I'll look into it.

tools/generator/templates/player/caml/interface.cc.jinja2
Dettorer: Old generator again, and I agree, I frowned when I copy/pasted it.

tools/generator/templates/player/caml/interface.cc.jinja2
```diff
@@ -0,0 +1,247 @@
...
+                                        {%- set comma = joiner(', ') -%}
+                                        {%- for arg_name, _, _ in arg_list -%}
+                                        {{ comma() }}value {{ arg_name }}
+                                        {%- endfor -%}
+                                       )
+{
+    {# TODO: This only works for functions with 5 parameters or less.
+     # To handle more, we must call CAMLparam5 once, then CAMLxparamN, grouping
+     # the arguments 5 by 5 maximum #}
+    CAMLparam{{ arg_list|length }}({# argument list #}
```
Dettorer, line 201: If the function doesn't have any parameter, we should be able to just call `CAMLparam0` and forget the `{%- set arg_list = func.fct_arg|caml_cxx_args %}`, I'll try that.

PR #101, opened 2020-03-02
generator: ocaml: make primitives handle an arbitrary number of args

seirl:


tools/generator/templates/player/caml/interface.cc.jinja2
```diff
@@ -195,15 +195,7 @@ extern "C" value ml_{{ func.fct_name }}({# argument list #}
...
-    {# TODO: This only works for functions with 5 parameters or less.
-     # To handle more, we must call CAMLparam5 once, then CAMLxparamN, grouping
-     # the arguments 5 by 5 maximum #}
-    CAMLparam{{ arg_list|length }}({# argument list #}
-                                   {%- set comma = joiner(', ') -%}
-                                   {%- for arg_name, _, _ in arg_list -%}
-                                   {{ comma() }}{{ arg_name }}
-                                   {%- endfor -%}
-                                  );
+    {{ arg_list|map('first')|caml_param_macro|indent }}
```
Dettorer, line 198: You should be able to use `func.fct_arg` here instead of `arg_list`, remove the line where it's set (https://github.com/prologin/stechec2/pull/101/files#diff-0ab55a5f02d7d1a0896ff290cce05feeR188) and remove the corresponding filter.

But to do that, you need to handle the empty arg list case either in the `caml_param_macro` filter or directly here in jinja (if there are no arguments, this line should be `CAMLparam0()`)

tools/generator/templates/player/caml/api.ml.jinja2
```diff
@@ -34,6 +34,8 @@ type {{ struct.str_name }} = {
 {% endfor %}
 {% for func in game.function %}
 {{ func.fct_summary|caml_comment(doc=True) }}
-{{ func|caml_signature(external=True) }} = "ml_{{ func.fct_name }}"
+{{ func|caml_signature(external=True) }} = {% if (func.fct_arg|length) > 5 -%}
+                                           "ml_bytecode_{{ func.fct_name }}"
+                                           {%- endif %} "ml_{{ func.fct_name }}"
```
Dettorer: Nit: this introduces an extra space between the `=` and the symbol name for regular functions.

Dettorer (changes requested):

Dettorer (approved):

PR #103, opened 2020-03-03
[WIP] Add a test for a user function with args & return value

remi-dupre:
I think C interface doesn't convert C types to C++ here: https://github.com/prologin/stechec2/blob/master/tools/generator/templates/player/c/champion.c.jinja2#L7-L11


remi-dupre:
An issue (https://github.com/prologin/stechec2/issues/105) was created as we may not be interested in addressing this issue for now.

Issue #42, opened 2019-05-19
Allow the rules to export stats of the games


seirl:
It would be cool if the rules had a way to export in their result some infos like:

- Number of turns
- Number of {action} done by each participant
- ...

This would allow us to show cool stats to the contestants, like:

- Shortest match
- Match with the greatest number of {action}
- ...

audebert:
Let's store the stats in files in the format described in: https://github.com/prometheus/docs/blob/master/content/docs/instrumenting/exposition_formats.md#text-format-details

Those files can then be consumed by https://github.com/prometheus/node_exporter#textfile-collector or by a simple parser to be displayed in the web interface.

Note: storing the match id and player id in metric labels will cause an explosion in metrics cardinality, we'll need to figure out how to aggregate those, which may be as simple as using https://prometheus.io/docs/prometheus/latest/configuration/recording_rules/#recording-rules

audebert:
Stats can also be emitted in logs, which can be processed by ElasticSearch and queried in Kibana.

juli0z:
Migrated to https://gitlab.com/prologin/tech/tools/stechec2/-/issues/42

PR #107, opened 2020-03-04
rules: introduce the --stats flag

audebert:
When used, the --stats FILE instructs stechec2-{server,client} to write
stats about the match to the specified file.

Currently collected stats:

* Player turn duration
* Total count of actions

Rules can add custom metrics to their local Rules::metrics_ instance
attribute.

For now only double-valued metrics can be collected.

Example stats output for a tictactoe server:

    - metrics:
      - actions_total: 6
    - player:
      - id: 0
      - name: anonymous
      - turn_duration:
        - 0.801743
        - 0.377152
        - 0.348772
    - player:
      - id: 1
      - name: anonymous
      - turn_duration:
        - 1.38622
        - 0.395657
        - 0.31234

Bug #42

### DIFF
--- a/src/lib/rules/player.hh
+++ b/src/lib/rules/player.hh
@@ -41,6 +41,7 @@ struct Player final
     int32_t score;
     uint32_t nb_timeout;
     std::string name;
+    std::vector<double> turn_duration_ms;
 };
 
 class Players final : public utils::IBufferizable
--- a/src/lib/rules/rules.cc
+++ b/src/lib/rules/rules.cc
@@ -3,8 +3,12 @@
 #include "rules.hh"
 
 #include <algorithm>
+#include <chrono>
+#include <fstream>
 #include <set>
 #include <unordered_set>
+
+DEFINE_string(stats, "", "YAML stats about the match output path.");
 
 namespace rules
 {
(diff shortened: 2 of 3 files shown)


seirl (changes requested):
- As discussed, JSON would be more suited for machine generated, structured data like this.
- Currently the dump and the game result can only be written by the server process (clients don't accept `--dump`). Is there any reason we don't follow that pattern here? It would simplify the logic a bit.

audebert:
> * As discussed, JSON would be more suited for machine generated, structured data like this.

As discussed, the rules will only export unstructured key=value metrics, which reduces the need to use JSON as the output format. On top of this there is currently no JSON helpers functions in the server/client code, only the rules have them. Adding JSON support would add complexity to the code.

> * Currently the dump and the game result can only be written by the server process (clients don't accept `--dump`). Is there any reason we don't follow that pattern here? It would simplify the logic a bit.

The clients have access to different pieces of information compared to the server, for example they will be able to count the cancels or gather precise timings for the champion turns, and we may also want to allow champions to generate their own stats in the future, which will only be accessible to the client.

seirl (approved):

Issue #38, opened 2018-05-14
generators: warn if yaml files contain reserved keywords in some languages


seirl:
It can't be perfect, but we can at least avoid the most common mistakes like this ("type", etc)

Shaac:
Solved by #109

PR #109, opened 2020-03-04
generator: check reserved keywords

Shaac:
#38

### DIFF
--- a/tools/generator/game.py
+++ b/tools/generator/game.py
@@ -7,6 +7,7 @@
 import re
 import typing
 import yaml
+from keyword import iskeyword
 
 
 def load_game(game_file):
@@ -113,6 +114,7 @@ class Game:
         for f in (self.game['function'] + self.game['user_function']):
             self.check_func(f)
         self.check_field_unicity()
+        self.check_reserved_keywords()
 
     def check_struct(self, struct):
         field_names = set()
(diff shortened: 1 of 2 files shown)


tools/generator/game.py
```diff
@@ -349,3 +401,47 @@ def explore(obj, schema, path):
         }
     ],
 }
+
+# Obtained with sorted(set(iorgen.parser_c.KEYWORDS +
+# iorgen.parser_cpp.KEYWORDS + iorgen.parser_csharp.KEYWORDS +
+# iorgen.parser_haskell.KEYWORDS + iorgen.parser_java.KEYWORDS +
+# iorgen.parser_php.KEYWORDS + iorgen.parser_rust.KEYWORDS))
+# Not that python keywords are not present, use keyword.iskeyword
```
remi-dupre:
```suggestion
# Note that python keywords are not present, use keyword.iskeyword
```

remi-dupre (commented):
Maybe we could benefit keeping keywords from each language separated? For instance it could help maintenance if a language is updating or if we want to stop supporting a language?

tools/generator/game.py
```diff
@@ -171,6 +173,56 @@ def check_field_unicity(self):
                         .format(origin_type, origin, name, def_type, def_name))
                 used_field_names[name] = (origin_type, origin)
 
+    def check_reserved_keywords(self):
+        '''Check that no identifier is in conflict with a reserved keyword'''
```
seirl, line 177: Seems like you could easily factor this with a nested function:

```
def check_ident(name, what):
    if conflicts_with_reserved_keyword(name):
        raise GameError("{} '{}': ...".format(what, name))
```

tools/generator/game.py
```diff
@@ -349,3 +401,47 @@ def explore(obj, schema, path):
...
+    'synchronized', 'template', 'then', 'this', 'thread_local', 'throw',
+    'throws', 'trait', 'transient', 'true', 'try', 'type', 'typedef', 'typeid',
+    'typename', 'typeof', 'uint', 'ulong', 'unchecked', 'union', 'unsafe',
+    'unset', 'unsigned', 'unsized', 'use', 'ushort', 'using', 'using static',
+    'var', 'virtual', 'void', 'volatile', 'wchar_t', 'where', 'while', 'xor',
+    'xor_eq', 'yield'
+]
+
+
+def conflicts_with_reserved_keyword(value: str) -> bool:
```
seirl: You could move that directly in the nested function above

Shaac:
> Maybe we could benefit keeping keywords from each language separated? For instance it could help maintenance if a language is updating or if we want to stop supporting a language?

Well it’s already separated lists inside iorgen. To update the list we just need to run the command in comment.

tools/generator/game.py
Shaac: fixed

tools/generator/game.py
Shaac: done

tools/generator/game.py
Shaac: done

seirl (approved):

PR #114, opened 2020-03-11
generator: load cst_type automatically

seirl:


tools/generator/game.py
```diff
@@ -44,11 +44,27 @@ def load(self):
...
+            str: 'string',
+            float: 'double',
+        }
+        for cst in self.game['constant']:
+            val_type = type(cst['cst_val'])
+            if val_type not in cst_type_mapping:
+                raise GameError("Constant '{}' has unsupported type: {}"
+                                .format(cst['cst_name'], val_type))
+            if 'cst_type' not in cst:
+                cst['cst_type'] = cst_type_mapping[val_type]
```
Shaac: Maybe add a check, `elif cst['cst_type'] != cst_type_mapping[val_type]: raise …`

Shaac (approved):

tools/generator/game.py
seirl: I thought about doing that, but I think we should maybe just deprecate the cst_type field in the yaml and ignore it completely

Shaac: The more reason to add at least a warning if the detected type is incorrect. If the user think its `2` will be interpreted as a float…

PR #116, opened 2020-03-12
Generate C/C++ closer to clang-format

Shaac:


seirl (approved):

tools/generator/templates/player/cxx/api.hh.jinja2
```diff
@@ -101,8 +103,9 @@ struct hash<{{ struct.str_name }}>
 {% endfor %}
 
 {# Demangle user functions #}
-extern "C" {
+extern "C"
+{
 {% for func in game.user_function %}
-extern "C" {{ func|cxx_prototype }};
+    {{ func|cxx_prototype }};
```
seirl, line 109: Do we indent externs now?

Shaac: Well as I said on IRC the other day, the option on the clang format to not indent is ignored, because `BreakBeforeBraces` needs to be set to custom, and the we have to set all the behavior for `Allman` by hand…

PR #119, opened 2020-04-06
generator: test: check that all tests are called + add missing ones

seirl:


tools/generator/test/languages/tester.cc
```diff
@@ -99,6 +130,7 @@ extern "C"
     {
         assert(s.field_i == 42);
         assert(s.field_bool);
```
Shaac, line 132: we should check that s.field_double in near 42.42

tools/generator/test/languages/champion.cs
```diff
@@ -50,18 +51,27 @@ public void Test()
       for (int i = 0; i < 7; ++i)
         Assert(v[i] == i + 1);
 
+      bool[] ba = {true, false, false, true, false, false, true, false, false};
+      ba = Api.ReturnsNot(ba);
+      for (int i = 0; i < 9; ++i) {
+          // XFAIL: the C# generator doesn't handle arrays of booleans
```
Shaac, line 57: Is there an issue for this?

Shaac (approved):

PR #121, opened 2020-04-07
Use shared pointer to the gamestate in the API

remi-dupre:
When using `unique_ptr`, we are not allowed to materialize two APIs that interact with the same game state. This is painful when writing tests as it makes it harder to simulate proper interactions between the players.

For examples, in tests for the past 4 years, `test-helpers.hh` gives access to two players, playing in two separate worlds:

 - (2019) https://github.com/prologin/prologin2019/blob/master/src/tests/test-helpers.hh#L104-L109
 - (2018) https://github.com/prologin/prologin2018/blob/master/src/tests/test-helpers.hh#L113-L118
 - (2017) https://github.com/prologin/prologin2017/blob/master/src/tests/test-helpers.hh#L51-L56
 - (2016) https://github.com/prologin/prologin2016/blob/master/src/tests/test-helpers.hh#L67-L71

Also note that this is breaking change, if this change seems fine I'll need to update code of past editions.

### DIFF
--- a/games/plusminus/src/api.cc
+++ b/games/plusminus/src/api.cc
@@ -9,7 +9,7 @@
 // global used in interface.cc
 Api* api;
 
-Api::Api(std::unique_ptr<GameState> game_state,
+Api::Api(std::shared_ptr<GameState> game_state,
          std::shared_ptr<rules::Player> player)
     : rules::Api<GameState, error>(std::move(game_state), player)
 {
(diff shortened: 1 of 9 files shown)


remi-dupre:
As discussed with @audebert on IRC, this is an unwise idea as the GameState is suposed to be an internal implementation detail of the API. If several API's are created they must hold with different gamestates that must be synced with similar mechanisms to what stechec server is doing.

PR #122, opened 2020-04-07
Augment the number of dumps and give them timestamp

fnareoh:
The goal is to add a notion of timestamp to the creation of a dump,
that would describe the round, the player and the action if there are
some. It comes with doing dumps much more often.

The use of a map is only a proof of concept to launch some discussion,
as it gives quite a lot of flexibility for what the timestamp can
represent.


juli0z:
Migrated to https://gitlab.com/prologin/tech/tools/stechec2/-/merge_requests/122

PR #125, opened 2020-04-11
generator: C#: handle arrays of arbitrary native types

### LINKED ISSUE #124: C# boolean arrays do not work
This test in particular always returns False: https://github.com/prologin/stechec2/blob/master/tools/generator/test/languages/champion.cs#L57-L58

seirl:
Fix #124

### DIFF
--- a/tools/generator/templates/player/cs/interface.cc.jinja2
+++ b/tools/generator/templates/player/cs/interface.cc.jinja2
@@ -9,6 +9,7 @@
 #include <assert.h>
 #include <cstdlib>
 #include <cstring>
+#include <functional>
 #include <iostream>
 #include <string>
 #include <vector>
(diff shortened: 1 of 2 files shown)


tools/generator/templates/player/cs/interface.cc.jinja2
```diff
@@ -38,6 +39,32 @@ class CSharpInterface
...
+template <> \
+const std::function<MonoClass*()> MonoClassFactory<CxxType>::get_class = \
+    Factory;
+
+MAP_TYPE_MONOCLASS(int, mono_get_int32_class);
+MAP_TYPE_MONOCLASS(double, mono_get_double_class);
+MAP_TYPE_MONOCLASS(bool, mono_get_boolean_class);
+MAP_TYPE_MONOCLASS(std::string, mono_get_string_class);
+
+#undef MAP_TYPE_MONOCLASS
```
Dettorer: Could something like
```c++
// Usage: get_mono_class<double>() -> mono_get_double_class()
template <typename T>
MonoClass* get_mono_class()
{
    if constexpr (std::is_same_v<T, int>)
        return mono_get_int32_class();
    else if constexpr (std::is_same_v<T, double>)
        return mono_get_double_class();
    else if constexpr (std::is_same_v<T, bool>)
        return mono_get_boolean_class();
    else if constexpr (std::is_same_v<T, std::string>)
        return mono_get_string_class();
    else
        return T::__if_that_triggers_an_error_there_is_a_problem;
}
```
be enough? It's shorter and less convoluted.

I successfully compiled and launched a C# champion with this version. I also tried using this function with an unsuported type and it fails properly at compile-time when trying to compile the content of the else.

tools/generator/templates/player/cs/interface.cc.jinja2
```diff
@@ -127,11 +154,10 @@ template <typename CsType, typename CxxType>
 MonoArray* cxx2lang_array(const std::vector<CxxType>& in)
 {
     gint32 size = in.size();
-    // TODO: handle string/double types
-    MonoClass* mcKlass = mono_get_int32_class();
+    MonoClass* mcKlass = MonoClassFactory<CxxType>::get_class();
```
Dettorer: This line would then be
```c++
MonoClass* mcKlass = get_mono_class<CxxType>();
```

Dettorer (approved):

PR #127, opened 2020-04-11
generator: add test coverage for double arrays

seirl:


tools/generator/test/languages/tester.cc
```diff
@@ -126,6 +127,14 @@ extern "C"
         return l;
     }
 
+    std::vector<double> api_returns_inverse(std::vector<double> l)
+    {
+        func_called.returns_inverse += 1;
+        for (unsigned i = 0; i < l.size(); ++i)
+            l[i] = 1 / l[i];
```
zopieux, line 134: Just curious: what happens if users passes 0 as input?

zopieux (approved):

PR #129, opened 2020-04-12
generator: add string tests and fixes

seirl:
This adds tests for string parameters for all languages.


zopieux (commented):
Nice but what about strings contained in some container? Could reveal memory issues.

zopieux (approved):

PR #135, opened 2020-04-30
Improve OCaml generator

FardaleM:
This does some cosmetics change, few improvement and correct a bug with float struct.

### DIFF
--- a/tools/generator/filters/caml.py
+++ b/tools/generator/filters/caml.py
@@ -28,11 +28,14 @@ def caml_cxx_args(arg_list):
 
 @register_filter
 def caml_prototype(func) -> str:
-    arg_list = ' '.join(arg_name for arg_name, _, _ in func['fct_arg'])
+    arg_list = ' '.join("({}:{})".format(arg_name, caml_type(arg_type))
+                        for arg_name, arg_type, _ in func['fct_arg'])
     if arg_list == '':  # make unit the only parameter
         arg_list = '()'
 
-    return 'let {} {} ='.format(func['fct_name'], arg_list)
+    return 'let {} {} : {} ='.format(func['fct_name'],
+                                     arg_list,
+                                     caml_type(func['fct_ret_type']))
 
 
 @register_filter
--- a/tools/generator/filters/caml.py
+++ b/tools/generator/filters/caml.py
@@ -1,6 +1,6 @@
 import textwrap
 
-from . import register_filter
+from . import register_filter, register_test
 from .common import get_array_inner, is_array
 from .cxx import cxx_type
 
@@ -100,3 +100,12 @@ def caml_param_macro(value) -> str:
             ', '.join(chunk)
         ))
     return '\n'.join(lines)
+
+
+@register_test
+def caml_test_float_struct(struct) -> bool:
+    """
+    Test if the struct contain only float fields.
+    """
+    return all(type_value == "double"
+               for _, type_value, _ in struct["str_field"])
--- a/tools/generator/templates/player/caml/api.ml.jinja2
+++ b/tools/generator/templates/player/caml/api.ml.jinja2
@@ -1,7 +1,7 @@
 {# Constants #}
 {% for constant in game.constant %}
-{{ constant.cst_comment|caml_comment(doc=True) }}
 let {{ constant.cst_name|lower }} = {{ constant.cst_val }}
+{{ constant.cst_comment|caml_comment(doc=True) }}
 
 {% endfor %}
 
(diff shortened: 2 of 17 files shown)


tools/generator/templates/player/caml/interface.cc.jinja2
```diff
@@ -183,10 +183,17 @@ value cxx_to_caml<value, {{ struct.str_name }}>({{ struct.str_name }} in)
 {
     CAMLparam0();
     CAMLlocal1(out);
+    {%if struct.str_field|caml_test_float_struct %}
```
seirl: nit: missing space

tools/generator/filters/caml.py
```diff
@@ -113,3 +113,12 @@ def caml_param_macro(value) -> str:
             ', '.join(chunk)
         ))
     return '\n'.join(lines)
+
+
+@register_filter
```
seirl: This should be a test instead. Use `@register_test` like this:

```python
@register_test
def caml_full_float_struct(s) -> bool:
    [...]
```

and use it like this in the template:

```jinja2
{% if struct is caml_full_float_struct %}
```

Also, the interface is nicer if you take the struct in parameter directly, instead of str_fields.

tools/generator/test/languages/champion.ml
```diff
@@ -49,6 +49,7 @@ let test () =  (* Pose ton code ici *)
     send_me_42s { field_int = 42;
                   field_int_arr = times42 42;
                   field_str_arr = times42 simple };
+    send_me_double_struct { field_one = 42.42; field_two = 42.42 };
```
seirl, line 52: Are you willing to add this to all the other test champions? It doesn't look like a ton of work, and it would allow you to check that the function is correctly called in all the languages.

I know we should probably have a way to specify language-specific tests, but in the meantime doing that isn't really difficult.

tools/generator/templates/player/caml/api.ml.jinja2
```diff
@@ -23,6 +23,11 @@ type {{ struct.str_name }} = ({%- set product = joiner(' * ') -%}
                               {%- for _, field_type, _ in struct.str_field -%}
                               {{ product() }}{{ field_type|caml_type }}
                               {%- endfor %})
+(*{%- set product = joiner(' , ') -%}
```
seirl: "product"?

tools/generator/templates/player/caml/champion.ml.jinja2
```diff
@@ -5,24 +5,20 @@ open Api
...
-{{ func.fct_summary|caml_comment }}
 {{ func|caml_prototype }}
-  begin
+{{ func.fct_summary|caml_comment(doc=True) }}
     (* TODO *)
 
     (* Pour s'assurer que les sorties s'affichent *)
     flush stderr;
-    flush stdout
+    flush stdout{{ func.fct_ret_type|caml_default_value }}
```
seirl: I don't think I like having a default value here. The only benefit is that it compiles from the start, but these functions should be changed anyway to remove the buggy default values, so why put them in the first place?

Also, it will only work for functions that don't return a special type, and it already worked for unit functions, which are the only ones we really support right now.

seirl (changes requested):
Overall good changes, I have some minor issues with two commits. Thanks for the PR!

tools/generator/templates/player/caml/api.ml.jinja2
```diff
@@ -23,6 +23,11 @@ type {{ struct.str_name }} = ({%- set product = joiner(' * ') -%}
                               {%- for _, field_type, _ in struct.str_field -%}
                               {{ product() }}{{ field_type|caml_type }}
                               {%- endfor %})
+(*{%- set comma = joiner(', ') -%}
```
seirl: Missing spaces after `(*` and before `*)`

seirl (changes requested):
Could you just revert your change of documentation order after the functions? https://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html puts the documentation before the signatures, and it's more consistent with the other languages.

Other than that and another nitpick, LGTM

FardaleM:
> Could you just revert your change of documentation order after the functions? https://caml.inria.fr/pub/docs/manual-ocaml/ocamldoc.html puts the documentation before the signatures, and it's more consistent with the other languages.
I don't see where ocamldoc puts documentation before signatures. If you speak about examples in the doc, there are comment before and after. Actually the rule to associate a comment with a element is simpler if the comment is bellow the element which can prevent future problem.

seirl (approved):

seirl:
Thanks!

PR #138, opened 2020-05-05
rules: Action: Add a virtual method dump_json

fnareoh:
Add an optional virtual method to output the json dump for an action, in
order to make the dump process more elegant for the rules that use this feature.

This way for the list of actions to dump you can just call `action.dump_json(gamestate)`.

### DIFF
--- a/tools/generator/templates/rules/src/action_template.hh.jinja2
+++ b/tools/generator/templates/rules/src/action_template.hh.jinja2
@@ -23,7 +23,7 @@ public:
 
     int check(const GameState& st) const override;
     void apply_on(GameState* st) const override;
-    void dump_json(const GameState& st) const;
+    void dump_json(const GameState& st, std::ostream& ss) const;
 
     void handle_buffer(utils::Buffer& buf) override
     {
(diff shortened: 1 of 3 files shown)


tools/generator/templates/rules/src/action_template.hh.jinja2
```diff
@@ -23,6 +23,7 @@ public:
 
     int check(const GameState& st) const override;
     void apply_on(GameState* st) const override;
+    void dump_json(const GameState& st, std::ostream& ss) const;
```
multun, line 26: I think I'd be slightly better with an `override`

multun (approved):
LGTM

seirl:
Hmm, why don't we reuse handle_buffer instead of having to write that function manually?

seirl (changes requested):
As explained above, this should use handle_buffer instead of reimplementing twice the serialization of each action.

juli0z:
Migrated to https://gitlab.com/prologin/tech/tools/stechec2/-/merge_requests/138

PR #142, opened 2020-05-10
generator: python: flush stdout and stderr after each function call

seirl:


tools/generator/templates/player/python/interface.cc.jinja2
```diff
@@ -311,6 +311,30 @@ static void _init_python()
...
+        if (tmp == nullptr)
+            PyErr_WriteUnraisable(fout);
+        else
+            Py_DECREF(tmp);
+    }
+
+    if (ferr != nullptr && ferr != Py_None) {
+        tmp = PyObject_CallMethod(ferr, "flush", "");
+        if (tmp == nullptr)
+            PyErr_Clear();
```
zopieux, line 332: Why the different error handing here? Because the error would go to stderr so it makes no sense to try to write to stderr if we couldn't flush?

zopieux (approved):
TBH I don't know the Python C API enough for my review to be relevant.

tools/generator/templates/player/python/interface.cc.jinja2
seirl: idk, this whole function is literally copy pasted from CPython.

PR #146, opened 2020-08-03
generator: fix tarball generation for some languages

alarsyo:


remi-dupre (approved):

alarsyo:
I've added a check to make sure we don't miss this next time :)

seirl (approved):
Nice! Good call for adding the make tar test.

PR #155, opened 2021-06-21
php: fix Makefile's php-config call

lportemo:


seirl:
We should probably do something like this to get the ldflags: https://github.com/weechat/weechat/blob/master/configure.ac#L862-L877

juli0z:
Migrated to https://gitlab.com/prologin/tech/tools/stechec2/-/merge_requests/155

PR #164, opened 2021-07-13
Test tuples in generated code

### LINKED ISSUE #161: Generator: add unit tests for tuples
The unit tests in https://github.com/prologin/stechec2/tree/master/tools/generator/test never test tuples, which leads to hard to catch issues like #160. They should be properly tested in all the languages that support it.

A lot of languages don't have tuple handling. Either we run useless tests on them that would be redundant with the struct tests, or we make a flag or something to run these tests only on the relevant languages.

LeSeulArtichaut:
Helps #161, cc @seirl

### DIFF
--- a/tools/generator/test/games/test.yml
+++ b/tools/generator/test/games/test.yml
@@ -33,6 +33,13 @@ struct:
       - [field_double, double, "Double field"]
       - [field_string, string, "String field"]
   -
+    str_name: simple_tuple
+    str_summary: Simple tuple structure.
+    str_tuple: yes
+    str_field:
+      - [field_0, int, "Tuple first field"]
+      - [field_1, bool, "Tuple second field"]
+  -
     str_name: struct_with_array
     str_summary: Structure which contains an array.
     str_tuple: no
@@ -54,13 +61,6 @@ struct:
     str_field:
       - [field_one, double, "First double"]
       - [field_two, double, "First double"]
-  -
-    str_name: simple_tuple_struct
-    str_summary: Simple tuple-like structure.
-    str_tuple: yes
-    str_field:
-      - [field_0, int, "Tuple first field"]
-      - [field_1, bool, "Tuple second field"]
 
 function:
   -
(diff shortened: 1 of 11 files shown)


seirl (commented):
Great! Two comments:

- we also need to check more complex types: tuples containing arrays/structs and arrays/structs containing tuples. I know there's an annoying combinatorial explosion here, but it's necessary unfortunately, we catch a lot of issues with tests like that. If you don't want to do that in this PR, just tell me, I'll merge that for now and update the issue progress.
- i'm surprised that you didn't use a tuple in the haskell test. There's specific code in the haskell generators to handle tuples. Does it just not work?

LeSeulArtichaut:
> i'm surprised that you didn't use a tuple in the haskell test. There's specific code in the haskell generators to handle tuples. Does it just not work?

The generated definition is this:
```haskell
data Simple_tuple_struct = Simple_tuple_struct Int Bool
```
As I understand it it's the same as a tuple in that it is a product type, but it's a different type so using a tuple instead will result in a type mismatch. If we wanted to use tuples instead of creating a new type, we should probably declare it as
```haskell
type Simple_type_struct = (Int, Bool)
```

On a similar note, I'm surprised that the Rust API doesn't generate either a tuple-like struct, i.e. `struct SimpleTupleStruct(i32, bool)`, or a `type SimpleTupleStruct = (i32, bool)` type alias to be used in signatures.

LeSeulArtichaut:
@seirl I'll be unavailable in the next few days, I think it'll be easier to add test cases for arrays and structs in a followup PR. I changed the PR description to not close #161.

EDIT: I'm back now :)

tools/generator/test/games/test.yml
```diff
@@ -198,6 +214,24 @@ function:
...
+    fct_arg:
+      - [s, simple_tuple, the tuple]
+  -
+    fct_name: send_me_struct_with_struct
+    fct_summary: Send me this struct with the usual test values for these fields
+    fct_ret_type: void
+    fct_arg:
+      - [s, struct_with_struct, the struct]
+  -
+    fct_name: send_me_tuple_with_struct
```
seirl, line 231: Right now you never check that `send_me_struct_with_struct` and `send_me_tuple_with_struct` are called.

LeSeulArtichaut: My bad, I forgot. I also need to add arrays into the mix for this PR to be ready

seirl (approved):

seirl:
Awesome work!

PR #165, opened 2021-07-18
Generator: minor Rust enhancements

LeSeulArtichaut:
- Use type aliases for tuples: for example, instead of using `(i32, i32)`, declare a type alias `type Position = (i32, i32)` and use it in the API. This makes the API more readable (functions take e.g. `pos: Position` instead of `pos: (i32, i32)`), and allows the type to have a rustdoc-generated documentation page where we can present the summary of the type and explanations for the fields.
- Document structs and enums: use the `summary` declared in the configuration to generate a doc-comment for structs and enums. This is already done in other languages, so not having them in Rust was probably an oversight;
- Code cleanups

:warning: This is a breaking change for tuples containing non-`Copy` fields. Say we have a `(String,)` tuple. Previously, we generated `(&String,)` in function arguments, but now we take the equivalent of `&(String,)`. This change however makes tuples more consistant with non-tuple structs

### DIFF
--- a/tools/generator/filters/rust.py
+++ b/tools/generator/filters/rust.py
@@ -87,22 +87,26 @@ def rust_api_output_type(ctx, value: str, api_mod_path='') -> str:
         'void': '()',
     }
 
-    as_struct = ctx['game'].get_struct(value)
-
     if is_array(value):
         return 'Vec<{}>'.format(
             rust_api_output_type(ctx, get_array_inner(value), api_mod_path)
         )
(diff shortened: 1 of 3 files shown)


seirl (approved):
Looks great! Could you just rebase on master so that we can check that the tests properly cover your new tuple function?

LeSeulArtichaut:
Oh, right, one test broke because of the breaking change I mentionned in the PR description. Fixing...

PR #166, opened 2021-07-19
CI: do a full reinstall of PHP to avoid package conflicts

LeSeulArtichaut:
The pre-installed version of PHP in GitHub Actions is from [DEB.SURY.ORG](https://deb.sury.org/), but `libphp-embed` is not available there, so to me it looks like we need to uninstall PHP and reinstall it from the official APT repos.

Fixes the recent CI errors. Alternative to #163, cc @seirl


seirl (approved):
OMG thanks this was driving me insane

LeSeulArtichaut:
I get you, I spent way too much time on this

PR #167, opened 2021-07-21
Fix C FFI type for arrays of tuples in Haskell

LeSeulArtichaut:
To handle transferring arrays to the C FFI, Haskell has a `CXXX_array` type for each used array type: https://github.com/prologin/stechec2/blob/49713208bd2ed3de9859c247736776544f483e9b/tools/generator/templates/player/haskell/CApi.hsc.jinja2#L194-L195
The `haskell_c_type` filter returns this type for arrays: https://github.com/prologin/stechec2/blob/49713208bd2ed3de9859c247736776544f483e9b/tools/generator/filters/haskell.py#L19-L23

However for tuples the templates used `haskell_type` instead, leading to invalid generated code like `C[XXX]`.


seirl:
Why did that fall through the tests? Should we add a new test case to cover this?

LeSeulArtichaut:
I couldn't add a test for arrays in tuples because of this bug and a problem in the Rust generator I'm still working on. Once they are fixed I have a test case ready.

seirl (approved):

PR #168, opened 2021-08-14
Test arrays in tuples

LeSeulArtichaut:
Continuation of #164, cc @seirl


seirl (approved):
Awesome, thanks a lot!

Issue #169, opened 2021-08-14
Generator: support constants of arbitrary types


LeSeulArtichaut:
Currently constants can only hold a value of type `int` or `double`. It might be useful to be able to have constants of any type, including the primitive types (`int`, `double`, `bool`), strings, arrays, structs and enums.

seirl:
Hmm, I'm pretty sure this won't work for C where constants are implemented as preprocessor macros.

LeSeulArtichaut:
Definitely not a C expert but I've tested it and it looks like you can have struct initializers in preprocessor macros. This compiles:
```c
#include <stdbool.h>

#define SIMPLE { .field_i = 42, .field_bool = true, .field_double = 42.42, .field_string = "TTY" }

/// Simple structure with nothing special.
typedef struct simple_struct
{
    int field_i; ///< Integer field
    bool field_bool; ///< Bool field
    double field_double; ///< Double field
    char* field_string; ///< String field
} simple_struct;

int main()
{
    simple_struct simple = SIMPLE;
    return 0;
}
```

seirl:
You're right, that should work.

I guess we also need to settle on a format to specify "values" in the yaml for complex types, but that should be straightforward too/

LeSeulArtichaut:
Actually, array/struct initializers in macros might be the wrong way to go, because, since they are not expressions, you can't access one of their elements/fields:

```c
int i = SIMPLE.field_i;  // error
```

A solution might be to use global `const` variables? Or `#define` for primitives and `const` for structs and arrays?

> I guess we also need to settle on a format to specify "values" in the yaml for complex types, but that should be straightforward too

The only potential problem I can see is that there is no easy way to represent an "identifier" in YAML. If we want to allow enum variants to be used in constants, or if we want to be able to reference other constants, we need to be able to disambiguate between an identifier to another object in the game and a string literal. I'm not sure however that these constructs would be useful.

juli0z:
Migrated to https://gitlab.com/prologin/tech/tools/stechec2/-/issues/169

Issue #159, opened 2021-07-12
generator: more idiomatic language APIs for errors and optional values


LeSeulArtichaut:
In this year's Prologin finals, I had discussions with fellow rustaceans on how the generated Rust API felt clumsy and unidiomatic at times. We concluded that it would be great if the API could better match the language's idioms and common patterns more.

## The problem

Here are two common patterns that would benefit greatly from a better API:

### Optional values

Sometimes, fetching some information through an API call can fail, e.g. if the provided input is invalid. In this case, games usually return a sentinel value that represents "no meaningful value", like `-1`. This might be idiomatic in C, but since optional values are literally everywhere in programming, modern (i.e. non-C) languages have better ways of representing optional values:
- Rust, OCaml and Haskell have `Option`, `option` and `Maybe` respectively
- Python, Java, C# and PHP all have some flavor of `null`

### Errors

Similarly, a champion might attempt to perform an invalid action, in which case the came usually responds with an error enumeration giving more details on why the attempted action was invalid. Once again, every language has its own error handling mechanisms (also see #89):
- Rust has `Result`, OCaml has `result`, Haskell has `Exception`*
- Python, Java, C# and PHP have exceptions*
- C++ programmers can't agree on how to handle errors, leave them alone

\* There is a caveat here: unhandled exceptions are propagated and can crash the program, and developers might be forced to write code to handle them (e.g. Java). Competitive programming is not the place for proper error handling and beautiful error messages. There is a trade-off here: having errors crash the program is good for debugging (no error goes unnoticed), but bad for tournaments where you want to continue the match no matter what, and bad for writing quick and dirty code.

## Proposed solution

The API is built on top of a C FFI layer, which can hardly be changed as it needs to be common to all languages. However, nothing prevents us from building language-specific abstractions on top of the FFI.
I think the two cases presented in this issue are fundamental and deserve special treatment. If we consider this to be true, then we can integrate them in the YAML config files used for specifying a game's API.

### Optional values

Functions can mark a function's return value as optional by specifying the sentinel value in a `fct_ret_none_val` field. This tells the generator to generate code to transform the sentinel into the langage's idiomatic way of representing missing values, like `None` in Rust or `null` in Java.

### Errors

Errors are specified in a new top-level list `error` (same level as `enum` or `function`), which has fields `err_name`, `err_summary`, `err_kind`. `err_kind` follows the same format as `enum_field`.

[...]

PR #170, opened 2021-08-14
Allow/fix strings in constants

LeSeulArtichaut:
Currently the YAML schema enforces that constants can't be strings, but `load_constants` has code to handle string values. This fixes the inconsistency by allowing strings to be used in constants.

### DIFF
--- a/tools/generator/game.py
+++ b/tools/generator/game.py
@@ -71,6 +71,10 @@ class Game:
                     "Constant '{}' type '{}' does not match value '{}'"
                     .format(cst['cst_name'], cst['cst_type'], cst_type)
                 )
+            if cst['cst_type'] == 'string':
+                # Quotation marks are removed when YAML is parsed,
+                # add them back
+                cst['cst_val'] = '"{}"'.format(cst['cst_val'])
 
     def load_base_types(self):
         '''
@@ -376,7 +380,8 @@ GAME_SCHEMA = {
     'constant': [
         {
             'cst_name': CONSTANT,
-            'cst_val': Union(int, float),
+            'cst_val': Union(int, str, float),
+            'cst_type': O(re.compile('(int|string|double)')),
             'cst_comment': str,
         }
     ],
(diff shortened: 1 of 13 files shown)


tools/generator/game.py
```diff
@@ -376,7 +380,8 @@ def explore(obj, schema, path):
     'constant': [
         {
             'cst_name': CONSTANT,
-            'cst_val': Union(int, float),
+            'cst_val': Union(int, str, float),
+            'cst_type': O(re.compile('(int|string|double)')),
```
seirl, line 384: I don't like having a cst_type field because it's redundant with the type of the value in the yaml. I thought we removed that field, is it still used?

LeSeulArtichaut: AFAICT the last (and only) time it was used besides tests was [in 2016](https://github.com/prologin/prologin2016/blob/38fee9eac97818b1002d0579d3e4d9172c8dab3d/prologin2016.yml#L49) and there's only one constant that uses it. It could be necessary to implement #169, but I guess that since enum and struct field names must be unique, we could easily infer the type based on that.

LeSeulArtichaut:
@seirl any feedback on https://github.com/prologin/stechec2/pull/170#discussion_r688993200 would be great

seirl:
Sorry if this is going slowly, I don't have the bandwidth to work on Prologin right now. I should have more time for this around mid-november!

LeSeulArtichaut:
@seirl Any chance you could take a look at this?

Also, if you have time for it, I'd love to get in touch to discuss #169 (and #159 more generally), for example on Discord. Let me know what you prefer.

seirl:
Brainstorming session: it would actually be more consistent to keep the `cst_type` field. Using only the first commit should probably be enough.

LeSeulArtichaut:
Removed the last commit, this should be ready to merge.

seirl (approved):
Looks good. Is it just me or are the constants not tested at all in many languages?

PR #171, opened 2021-10-17
Add generator `--check` flag

### LINKED ISSUE #162: Generator: command to check that a .yml compiles in all languages
We don't have a way to quickly check that a given .yml generates code that compiles properly. It always should, but a command like that would be useful to investigate errors in the generators.

Usage example:

```
$ stechec2-generator playercheck ./prologin2018.yml
c... OK
cxx... OK
cs... OK
java... OK
[...]
```

LeSeulArtichaut:
Closes #162.

### DIFF
--- a/tools/generator/player.py
+++ b/tools/generator/player.py
@@ -7,6 +7,7 @@ import os
 import subprocess
 from pathlib import Path
 from sys import stderr
+from typing import Optional
 
 from .generator import Generator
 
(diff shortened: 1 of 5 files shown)


seirl (changes requested):
I don't like having the compiling/checking code as part of the gen_lang function. I'd much rather have a function that can check an already generated directory. It would be more modular/reusable that way.

Then we could have a function `checkplayer` that only takes a `.yml` file, generates the player skeleton in a temporary directory, checks that it compiles, then removes the directory.

LeSeulArtichaut:
Reorganized the code a bit, does that look better?

tools/generator/test/test_languages.py
```diff
@@ -112,3 +116,23 @@ def compile_language(self, language, champion_file_name):
             )
         except subprocess.CalledProcessError as e:
             self.fail("Tester compilation failed with output:\n" + e.output)
+
+
+class CheckLanguages(unittest.TestCase):
```
seirl: It took me a while to understand what this class is doing, and how it's different from TestLanguages above. Rename to `TestExampleGames` and `def test_example_games(self)`?

tools/generator/test/test_languages.py
```diff
@@ -8,14 +8,18 @@
...
 import unittest
+from sys import stderr
 
-from generator.test.utils import generate_player
+from generator.game import load_game
+from generator.player import check_player
+from generator.test.utils import generate_player, load_test_game
 
 
 class TestLanguages(unittest.TestCase):
```
seirl, line 18: Could the `compile_language()` function here be refactored with the `check_compile()` function in player.py? Seems like there's some common behavior? (Tell me if you think it's not worth it.)

tools/generator/player.py
```diff
@@ -32,32 +67,50 @@ def gen_lang(lang: str, *, files, symlinks) -> None:
...
-             symlinks=['api.py', 'interface.cc', 'Makefile-python'])
-    gen_lang('rust',
-             files=['Makefile', 'champion.rs'],
-             symlinks=['Cargo.toml', 'Makefile-rust', 'api.rs', 'ffi.rs',
-                       'api.h', 'interface.cc'])
+    for lang, files in LANGUAGES.items():
+        gen_lang(lang, **files)
+
+
+def check_compile(lang_dir: Path, lang: str, *, cleanup=True) -> bool:
```
seirl: Can we also test that `make tar` works?

tools/generator/player.py
```diff
@@ -32,32 +67,50 @@ def gen_lang(lang: str, *, files, symlinks) -> None:
...
+            capture_output=True, cwd=lang_dir,
+        )
+    except subprocess.CalledProcessError as e:
+        success = False
+        eprint("FAILED")
+        eprint(e.stderr)
+    else:
+        eprint("OK")
+
+    if cleanup:
```
seirl: I don't think this is needed, if you call `--check` on an already existing dir it's probably because you want to debug with the target output files. Otherwise you should use a `checkplayer` command that generates everything in a temporary directory, and here it will all get wiped out when the command ends anyway.

seirl (changes requested):
Great overall, just a few additional comments.

tools/generator/test/test_languages.py
LeSeulArtichaut: Made an attempt to refactor, I guess it gives slightly worse output (no distinction for a failure in `make` or `make tar`)

seirl: Maybe `check_compile()` and `check_tar()`?

LeSeulArtichaut:
Resolved the last review comment. Let me know if you want me to squash my commits

seirl (approved):

PR #172, opened 2022-04-18
packaging: package stechec2 and games in nix

lportemo:


flake.nix
```diff
@@ -0,0 +1,25 @@
...
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixos-21.11";
+  };
+
+  outputs = { self, nixpkgs }: 
+  let
+    pkgs = import nixpkgs { system = "x86_64-linux"; };
+  in
+
+  {
```
delroth: forAllSystems (or use flake-utils). No reason to have this flake be x86_64-linux specific.

nix/tictactoe.nix
```diff
@@ -0,0 +1,44 @@
+{ self, pkgs, ... }:
+
+
+let
+  stechec2-src = fetchGit {
```
delroth: I'd probably just have the stechec2 derivation expose its source instead (via passthru), hardcoding this seems fragile.

nix/tictactoe.nix
```diff
@@ -0,0 +1,44 @@
...
+    cppzmq
+    gcc
+    gflags
+    gcovr
+    pkg-config
+    wafHook
+  ];
+
+  preConfigure = ''
+    wafConfigureFlags="--with-games=tictactoe --games-only"
```
delroth: Why do that as a shell script in preConfigure instead of just `wafConfigureFlags = [ "--with-games=tictactoe" "--games-only" ];` ?

nix/tictactoe.nix
```diff
@@ -0,0 +1,44 @@
...
+  };
+in
+
+pkgs.stdenv.mkDerivation {
+  name = "tictactoe";
+  version = "1.0";
+
+  src = stechecWithGame;
+
+  buildInputs = with pkgs; [
```
delroth: pkg-config and wafHook should likely be nativeBuildInputs. gtest/gcovr likely checkInputs. Not sure why gcc needs to be in there, it's part of the stdenv, if it's actually needed in buildInputs that's likely hiding a serious bug (e.g. using the wrong compiler when building in certain environments).

(Similar comment for the deps list in the main stechec2 derivation -- these two should be deduped somehow.)

nix/tictactoe.nix
```diff
@@ -0,0 +1,44 @@
+{ self, pkgs, ... }:
```
delroth: `self` is not used.

flake.nix
```diff
@@ -0,0 +1,25 @@
+{
+  description = "A very basic flake";
```
delroth: FIXME

nix/stechec2.nix
```diff
@@ -27,6 +27,12 @@ let
       gcovr
     ];
 
+    passthru.deps = {
+      nativeBuildInputs = nativeBuildInputs;
```
zopieux: I think you can use `inherit` here.

juli0z:
Migrated to https://gitlab.com/prologin/tech/tools/stechec2/-/merge_requests/172

PR #176, opened 2022-05-30
generator: jinja 3.1 support

Shaac:
jinja 3.0 deprecated `contextfilter`, 3.1 removed it.

This PR uses the new name, while keeping compatibility for jinja < 3.

Tested with jinja 3.1.2 and 2.11.3


seirl (approved):
LGTM, thanks!

PR #177, opened 2022-05-31
generator/python: add docstrings and type hints

### LINKED ISSUE #173: Python generators should generate a .pyi file to enable autocompletion

Shaac:
Generates the .pyi file for the API functions. They come with type hints and docstring.

I also changed the comments describing enums and structs into docstrings so that IDE can see them.

Fixes #173

### DIFF
--- a/tools/generator/filters/python.py
+++ b/tools/generator/filters/python.py
@@ -11,7 +11,7 @@ from .cxx import cxx_type
 
 @register_filter
 @pass_context
-def python_type(ctx, val: str) -> str:
+def python_type(ctx, val: str, tuple_alias: bool = True) -> str:
     base_types = {
         'void': 'None',
         'int': 'int',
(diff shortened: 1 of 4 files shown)


tools/generator/filters/python.py
```diff
@@ -60,3 +81,30 @@ def python_to_cxx(value: str) -> str:
         )
     else:
         return 'python_to_cxx<PyObject*, {}>'.format(cxx_type(value))
+
+
+@register_filter
+def get_python_import(functions) -> str:
```
seirl, line 87: You should make this a function with pass_context instead. Then you can just look up the function dict in the context. It's cleaner and if other types start requiring imports, we won't have to change the templates.

Shaac: Well the point here is that we only need the imports for what is declared in `functions`. Because `_api.pyi` only declares `functions`. If other types need imports in `_api.pyi`, it’s a good think that we need to change the templates.

Keeping the function as is, we don’t have to modify it to reuse it in an other template, that would only need the imports for one single function.

But in both cases it’s speculation of what would be other uses. I don’t think it matters much.

tools/generator/filters/python.py
```diff
@@ -11,7 +11,7 @@
 
 @register_filter
 @pass_context
-def python_type(ctx, val: str) -> str:
+def python_type(ctx, val: str, tuple_alias: bool = True) -> str:
```
seirl, line 14: Could you add a short documentation to the function? It's hard to understand what tuple_alias means here

seirl (approved):
LGTM aside from a minor documentation issue

juli0z:
Migrated to https://gitlab.com/prologin/tech/tools/stechec2/-/merge_requests/177

PR #178, opened 2022-05-31
generator: check func name / fields conflicts

Shaac:
In Haskell, struct fields conflict with function names.

There is no real reason to check conflicts with enums currently, but I feel it's safer that way.


seirl:
Obsoleted by #179

Issue #143, opened 2020-07-25
rules generator assumes that actions error type is called `error`


pguenezan:
Redefine `afficher_erreur` in `interface.cc` and in `api.{cc,hh}`, `erreur` is rename to `error`

remi-dupre:
More generally, I think we want to map errors to a user-defined type.

[This is not the first time that it seems that errors could use a special treatment](https://github.com/prologin/stechec2/issues/89), so maybe it would be a good idea to implement such a thing.

A simple way to handle it would be to add a boolean field `is_error_type` to enum definitions, however it doesn't explicit the fact that there must be one and exactly one of such type.

seirl:
> however it doesn't explicit the fact that there must be one and exactly one of such type.

Is that something we want to enforce? I don't see why we couldn't have multiple error types for different functions.

remi-dupre:
> Is that something we want to enforce? I don't see why we couldn't have multiple error types for different functions.

I'm thinking about the error type attached to the API, which will be returned by actions. This "special" error type is passed as a template parameter to `rules::Api` [here](https://github.com/prologin/stechec2/blob/master/src/lib/rules/api.hh#L14).

seirl:
Yes, I know that it got recently specialized to always work that way, I'm just not sure that's a good approach. Conceptually there is really no reason why all API functions should return the same error type.

remi-dupre:
The logic around errors seems very short and quite easy to generalize indeed.

juli0z:
Migrated to https://gitlab.com/prologin/tech/tools/stechec2/-/issues/143

PR #181, opened 2022-06-07
MVP support for idiomatic errors

LeSeulArtichaut:
This PR implements support for idiomatic errors as discussed as part of #159.

### Design

Enumerations can be marked as `enum_error: yes` in configuration files. This tells `stechec` to treat the first variant of the enum as success ("OK") and all other variants as errors.

Marking an enum as an error changes its representation depending on the language:
- in Rust: `Result<(), Error>`
- in Haskell: `Either () Error`
- in OCaml: `(unit, error) result`
- in all other languages, the representation is unchanged compared to a regular enum

### DIFF
--- a/doc/rules.rst
+++ b/doc/rules.rst
@@ -145,6 +145,10 @@ An enumeration is described by three fields:
 
 * ``enum_name``: the name of the enum.
 * ``enum_summary``: a short documentation of the enum.
+* ``enum_error`` (yes/no): if set to "yes", the enum will be interpreted as
+  an error and enable some sugar in the languages that support it. The first
+  variant of the enum will represent success and all other variants represent
+  errors.
 * ``enum_field``: a list of the different fields of the enum. Each field is a
   list containing:
 
@@ -162,6 +166,7 @@ We can write it quickly::
       enum_summary:
         "Enumeration containing all possible error types that can be returned
         by action functions"
+      enum_error: yes
       enum_field:
         - [ok, "no error occurred"]
         - [out_of_bounds, "provided position is out of bounds"]
--- a/games/plusminus/champions/cxx/prologin.hh
+++ b/games/plusminus/champions/cxx/prologin.hh
@@ -15,7 +15,7 @@
 typedef enum error
 {
     OK,           /* <- no error occurred */
-    OUT_OF_BOUDS, /* <- guess is out of bounds */
+    OUT_OF_BOUNDS, /* <- guess is out of bounds */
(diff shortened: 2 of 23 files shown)


seirl:
Nice. In the issue we said that it would probably be better to have all enums potentially be errors, and then have the option in functions instead (like `fct_returns_error: true`). Could you explain why you decided to go back to the idea of having the flag on the enums instead? Would it make the code a lot more complex?

LeSeulArtichaut:
My bad, I forgot to post it somewhere.

The thing I realized when tinkering around is that since sugar happens at the type level, a flag on the enum makes implementation easier.
The annoying thing about the flag on the function is that of namespacing. Since the error is just a normal enum, it would be generated without sugar, and then you'd need to find a different name for the new error (without OK). And since error enums will probably not be used for any other purpose than errors, you get duplicated representations for virtually no benefit.

There is also the alternative of having errors as `throws` fields on functions, which would be more expressive, but that can be built on top of the sugar implemented in this PR. However it runs into namespacing problems as well and I'm not sure how to resolve them, though I haven't spent a lot of time thinking about it.

seirl:
Thanks, that makes sense. Could you also check that #143 can be properly fixed with this change? Not necessarily implement it yourself (although it would be great), but we just don't want to be stuck if it ends up not being sufficient to fix it.

LeSeulArtichaut:
I would say this PR doesn't change the status of #143, which is a "flaw" in the design of the rules code structure, while this PR just adds sugar to player APIs. The rules templates do use the same functions as the player generator, but they are unchanged in C++.

seirl:
Right, my question was more about whether the `enum_error: yes` field would be enough information to implement the fix in the C++ rules code, but it seems like it should.

LeSeulArtichaut:
You mean, we could enforce that there exists one and only one error for a game, and use that enum in the rules code ? Or is there another idea/fix that I'm missing? Merging all errors from all enums into a single one? I guess that discussion should take place in the issue.

LeSeulArtichaut:
Oh, I didn't understand the rules code properly. Yes, it looks like to me that this would suffice. I can try to implement it before merging this PR

LeSeulArtichaut:
Alright, I'll hold off from implementing a fix for #143. My C++ knowledge is lacking and the rules architecture is pretty complex.

juli0z:
Migrated to https://gitlab.com/prologin/tech/tools/stechec2/-/merge_requests/181

